PR #11, opened 2026-08-01
fix: code-review findings — token accounting, decompress, auto-update, dead code

ranxianglei:
Addresses all findings from the project review (including acp-kernel integration).

## Token counting (P0)
`estimateTokens` switched from `chars/4` to acp-kernel's `defaultCountTokens`. The old estimator diverged from the kernel's own counter — **Chinese under-counted 3.4×, code over-counted 2.8×** — so nudge / emergency-truncate thresholds fired on the wrong usage ratio. The kernel already calls `defaultCountTokens` internally every turn (9+ sites), so there is no new overhead.

**Performance verified** (the concern raised in review):

| input | ~tokens | `defaultCountTokens` |
|-------|---------|----------------------|
| 10 MB | 2.1M (max real window is ~2M) | 89 ms |
| 100 MB | 21M | 1.06 s |
| 500 MB | 104M | 8.4 s |

Realistic model windows (≤2M tokens ≈ 8 MB) cost ~70 ms once per turn — negligible next to the LLM call. Added `tests/tokens.test.ts` (CJK + coverage-skip cases).

## `/acp-decompress` command (P0)
Now performs a **real** decompression (`deactivateBlock` + `runtime.save` + preview), matching the decompress tool and the opencode-acp pattern. Previously it only read the summary via `core.decompress` (a pure lookup) and never mutated state.

## Auto-update opt-out (P1)
New `AdapterConfig.autoUpdate` (default `true`) and env `ACP_AUTO_UPDATE=0` disable all startup network calls.

## Safer auto-install (P1)
Dropped the silent host `package.json` rewrite; now runs scoped `npm install pai-acp@<ver>`, checks the npm exit code (only reports success on `0`), timeout 30s → 60s. No longer re-installs unrelated deps or falsely reports success on failure.

## Cleanup (P2)
- Removed dead `ACP_TAG_EXAMPLE` / `LT` / `GT` constants (system prompt content kept intact).
- Removed discarded `void estimateTokens(...)` in status-tool (`buildStatusReport` computes tokens itself).
- `update.ts`: removed redundant dynamic imports of `fs/promises` + `path`.
- `index.ts` reuses `runtime.stateFor` (now also returns `entries`).
- Pinned `acp-kernel` to exact `0.0.6` (0.x caret only locks `0.0.x`).

## Quality
typecheck ✅ · 20/20 tests ✅ · build ✅ (`dist/index.js` 126 KB)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Blocks: 3 active (3.7K summary, 15.2K original compressed)
 
 ### Auto-update
 
-On each Pi startup, pai-acp checks npm for a newer version and auto-installs it. No manual updates needed — just restart Pi.
+On each Pi startup, pai-acp checks npm for a newer version and auto-installs it. No manual updates needed — just restart Pi. To disable (no network calls on startup), set `autoUpdate: false` in the config or env `ACP_AUTO_UPDATE=0`.
 
 ### Compression philosophy in system prompt
 
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Blocks: 3 active (3.7K summary, 15.2K original compressed)
 
 ### Auto-update
 
-On each Pi startup, pai-acp checks npm for a newer version and auto-installs it. No manual updates needed — just restart Pi. To disable (no network calls on startup), set `autoUpdate: false` in the config or env `ACP_AUTO_UPDATE=0`.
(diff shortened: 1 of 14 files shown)


ranxianglei:
Superseded by #12 (merged). This branch was based on the pre-fix commit 4239962 (150K cap). All useful fixes were reapplied onto current master as #12 and merged cleanly. A direct diff vs master shows the only net changes here would be regressions from the old base: version 0.1.4→0.1.3, re-introducing the 150K cap (reverted in 2a8834a), and deleting tests/config.test.ts. No new value — closing.

PR #72, opened 2026-08-07
fix: surface compression warnings to the model

Tyan66666:
## Problem

The kernel's `applyCompression` returns `warnings: string[]` (types.d.ts:199 — *"host should surface these to the model"*), generated when soft-protected messages (recent N / last N tokens / last user message) are excluded from the requested compression range, e.g.:

```
Excluded 12 protected message(s) m00548...m00565 from compression range (recent/last-user zone).
```

But the adapter's `compress-tool.ts` destructured only `{ blocksCreated, tokensCompressed, errors }` — **the warnings were silently dropped**.

### Consequence: the model's mental map diverges from reality

Not data loss — a map error. The model believes it compressed `[A..Z]` into summary `S`; in reality only `[A..Y]` was folded and the protected `[Y+1..Z]` remains visible. When the model later references a message in the excluded segment:

1. It looks in `decompress(S)` → **not there** → comes up empty
2. The truth sits in the visible tail — but the model doesn't look there, because its map says that range no longer exists
3. Secondary effect: token accounting is off (believes the whole range was freed)

## Fix (adapter-level, 2 lines)

```ts
const { blocksCreated, tokensCompressed, errors, warnings } = applied.result;
// ...
if (warnings.length > 0) lines.push("⚠️ " + warnings.join("; "));
```

The kernel already did its part (produced the warnings); this fixes the adapter violating the kernel's contract by not surfacing them. No kernel changes needed.

## Boundary note

Hard-protected exclusions (`ALWAYS_PROTECTED_TOOLS = ["compress"]`) are **silent by design** — they are deterministic, expected rules, and the kernel does not emit warnings for them. ⚠️ fires only for soft-protection (recent-zone / last-user) trimming, which is genuinely unexpected from the model's perspective.

## Tests

1 new test in `tests/warnings-passthrough.test.ts` (soft-protection exclusion → ⚠️ line with `Excluded N protected message(s)` in the result). Full suite: **48 pass**, typecheck clean. Live-verified in the running Pi session:

```
▣ ACP | 13.5K → 5.1K tokens (~8.4K reclaimed, 5 blocks)
⚠️ Excluded 12 protected message(s) m00548, m00549, ... m00565 from compression range (recent/last-user zone).;
   Excluded 4 protected message(s) m00642, m00643, m00656, m00657 ...
```

### DIFF
--- a/src/compress-tool.ts
+++ b/src/compress-tool.ts
@@ -75,7 +75,7 @@ async function handleCompress(args: CompressArgs, runtime: AcpRuntime, ctx: Exte
     config,
   });
   await runtime.save(applied.state, ctx);
-  const { blocksCreated, tokensCompressed, errors } = applied.result;
+  const { blocksCreated, tokensCompressed, errors, warnings } = applied.result;
 
   const afterTokens = Math.max(0, beforeTokens - tokensCompressed);
 
@@ -95,6 +95,7 @@ async function handleCompress(args: CompressArgs, runtime: AcpRuntime, ctx: Exte
   });
 
   const lines = [`▣ ACP | ${formatK(beforeTokens)} → ${formatK(afterTokens)} tokens (~${formatK(tokensCompressed)} reclaimed, ${blocksCreated} block${blocksCreated > 1 ? "s" : ""})`];
+  if (warnings.length > 0) lines.push("⚠️ " + warnings.join("; "));
   if (errors.length > 0) lines.push("Errors: " + errors.join("; "));
   return lines.join("\n");
 }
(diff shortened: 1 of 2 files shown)


ranxianglei:
@Tyan66666 你好,想问下这个 PR 为什么关闭了?😅

我们这边刚做完 review:本地验证了 typecheck / 全部 48 个测试 / build 都通过,改动定位也很准确(就是把 kernel 已经生成的 `warnings` 字段从 `compress-tool` 透传给模型,维持压缩后"地图=现实"的不变量),正准备合并,结果发现已经被关了。

是因为遇到什么问题了吗?还是有别的考虑(比如打算重开一个、或者换个实现思路)?方便的话说一下,我们也好对齐。这个修复本身是有价值的,要是能合进来挺好的。

Tyan66666:
哦，没有别考虑，单纯就是你的 action 执行失败了。我看名字可能要换一下，就重新创建了一个，发现还是不行。

PR #77, opened 2026-08-07
fix(delegate): pass --tools whitelist to child process

Zeno2019:
## What

`buildChildArgs` was not passing the agent's tool whitelist to child Pi processes. A "read-only" reviewer child could receive edit/write tools — violating the role's access contract.

## Changes

- `src/delegate-tool.ts`: `buildChildArgs` looks up `AGENTS[args.agent].tools` and appends `--tools <whitelist>` to CLI args
- `tests/delegate-tool.test.ts`: +5 tests (reviewer/worker whitelists, arg ordering, unknown agent, model inheritance)

## Verification

- typecheck ✅ | build ✅ | 58/58 tests pass

### DIFF
--- a/src/delegate-tool.ts
+++ b/src/delegate-tool.ts
@@ -481,7 +481,7 @@ async function runDelegate(
   return formatSyncResult(args.agent, runId, args.task, result, file);
 }
 
-async function buildChildArgs(
+export async function buildChildArgs(
   args: DelegateArgs,
   rolePrompt: string,
   ctx: ExtensionContext,
@@ -493,6 +493,13 @@ async function buildChildArgs(
   await writeFile(promptFile, `${rolePrompt}\n\n---\n\nComplete the task below.`, "utf8");
 
   const cliArgs = ["-p", "--no-session", "--append-system-prompt", promptFile];
+
+  // Pass the role's tool whitelist to the child (--tools is Pi CLI native).
+  // read-only roles must not receive edit/write; also disables child extension tools.
+  const agentDef = AGENTS[args.agent];
+  if (agentDef?.tools) {
+    cliArgs.push("--tools", agentDef.tools);
+  }
 
   if (args.model && args.model.includes("/")) {
     const [providerId, ...rest] = args.model.split("/");
--- a/tests/delegate-tool.test.ts
+++ b/tests/delegate-tool.test.ts
@@ -1,0 +1,67 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { buildChildArgs } from "../src/delegate-tool.js";
+import type { ExtensionContext } from "@earendil-works/pi-coding-agent";
+
+/** Minimal ctx mock — buildChildArgs only reads ctx.model. */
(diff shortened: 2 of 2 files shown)


ranxianglei:
Thanks for this PR — the intent (making the delegated child process respect each role's tool whitelist) makes sense. But after digging into it, I think there are some deeper issues worth discussing before merging, because I'm not sure this actually achieves the access control it implies.

**1. The whitelist doesn't truly enforce read-only.**

Every read-only role (reviewer / researcher / planner / oracle) still gets `bash`. `bash` is a strict superset of `edit` / `write`:

```
echo "..." > file          # write
sed -i 's/old/new/' file   # edit
rm file                    # delete — even beyond edit/write
```

So "the reviewer can't edit/write" only holds for the *direct tool call*. A single shell command bypasses it completely. From an access-control standpoint the boundary is leaky.

**2. The dilemma this creates.**

The only way to make read-only *real* is to drop `bash` from those roles (keep just `read`, maybe `grep` / `find` / `ls`). But then a reviewer often *can't do its job* — searching the codebase, running tests/builds to validate the review, etc. all need the shell. So we're stuck between "not enforceable" and "can't work."

**3. Tool-list configuration gets complicated fast.**

As soon as we start curating per-role tool lists, a lot falls into scope:
- Our own context-management tools (`compress` / `decompress` / `search_context` / `acp_status`) — arguably the delegated child should also get these, otherwise it can't manage its own context under ACP.
- Third-party tools the user has installed and wants the child to use.
- Today the whitelist is hardcoded per role; supporting the above means making it *configurable*, and getting that configuration right is non-trivial.

**4. Precedent: `subagents`.**

I looked into `subagents` (another delegate-style extension) for comparison and hit the same wall: there's no clean way to make a delegated agent inherit the ACP compression pipeline. Either we inject/override the tool set, or we patch it at install time — neither is clean. `subagents` effectively can't support our context compression out of the box.

**5. Delegating the "main" task.**

There are also scenarios where a user delegates the *primary* task (not just a read-only review) to a child agent. If we now require per-role tool configuration — or push that config burden onto the user — the setup becomes heavy, which hurts the common case.

**My takeaway.**

I suspect this is less a flaw in this PR and more a fundamental tension in how Pi lets extensions sandbox/shape a child process: `--tools` is a *capability hint*, not a *capability boundary* (because `bash` exists), and there's no first-class mechanism in Pi to say "this role is truly read-only" or "this child should inherit these extension tools."

So I'd like to **open this up for discussion** rather than merge as-is. Some possible directions:

- **Accept as a soft guardrail:** merge this PR but clearly document that it only prevents *accidental/direct* edit/write and advertises the right toolset — it is *not* access control.
- **Push further:** define read-only roles without `bash` (+ grep/find/ls), accept the reduced capability, and work out the config story for ACP / third-party tools.
- **Reconsider the isolation unit:** whether delegate-style child processes are the right boundary at all vs. something Pi should provide natively.

Would love @ranxianglei's take too. Happy to help prototype whichever direction we land on.

PR #79, opened 2026-08-07
fix(delegate): pass --tools whitelist to child process

Zeno2019:
## What

`buildChildArgs` was not passing agent tool whitelists to child Pi processes. A read-only reviewer could receive edit/write tools.

## Changes

- `src/delegate-tool.ts`: `buildChildArgs` appends `--tools <whitelist>` from `AGENTS[args.agent].tools`
- `tests/delegate-tool.test.ts`: +5 tests

## Verification

typecheck ✅ | build ✅ | 58/58 tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,17 +86,21 @@ This has two practical implications:
 
 ### acp_delegate — clean-context delegation
 
-Hand a self-contained task to a fresh pi process running in a clean context. Five built-in roles, each with a tailored tool whitelist and system prompt:
+Hand a self-contained task to a fresh pi process running in a clean context. Five built-in roles, each with a system prompt and a **soft tool guardrail**:
 
 | Role | Tools | Best for |
 |------|-------|----------|
-| `reviewer` | read, bash | Read-only code review (bugs, risks, file:line) |
-| `researcher` | read, bash | Read-only codebase investigation |
+| `reviewer` | read, bash, grep, find, ls + ACP | Read-only code review (bugs, risks, file:line) |
+| `researcher` | read, bash, grep, find, ls + ACP | Read-only codebase investigation |
 | `worker` | read, edit, write, bash | Make code changes |
-| `planner` | read, bash | Analyze + propose a step-by-step plan |
-| `oracle` | read, bash | Answer questions / advise |
+| `planner` | read, bash, grep, find, ls + ACP | Analyze + propose a step-by-step plan |
+| `oracle` | read, bash, grep, find, ls + ACP | Answer questions / advise |
 
-The full delegate result is saved to a file (`/tmp/acp-delegate/<runId>.out`); the tool result and injected notification carry only the **task title + file path** (no preview) — use `read` for the details. This keeps the parent context lean.
+Read-only roles (reviewer, researcher, planner, oracle) receive a restricted tool allowlist (`read, bash, grep, find, ls`) plus ACP context tools (`compress, decompress, search_context, acp_status`) so they can manage their own context. This prevents accidental file modifications, but `bash` can bypass it - **it is a guardrail, not a security boundary**.
+
+Worker runs on Pi's full default toolset - no `--tools` allowlist is applied, so any loaded extension or custom tools (e.g. ACP, LSP, MCP) remain available. This keeps primary-task delegation fully capable. The `read, edit, write, bash` listing above reflects core tools only.
+
+The full delegate result is saved to a file (`/tmp/acp-delegate/<runId>.out`); the tool result and injected notification carry only the **task title + file path** (no preview) - use `read` for the details. This keeps the parent context lean.
 
 - **Interactive (TUI) & RPC modes**: `async:true` (default) runs the child in the background; a short completion notification is injected into the chat when it finishes.
 - **Print / JSON modes** (`pi -p`, SDK): `async:true` auto-downgrades to **synchronous** — the result returns as the tool result in the same turn (the parent exits after one turn, so background injection would be lost).
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -85,17 +85,21 @@ billion-context 通过拦截 Pi 的 `context` 事件接管上下文管理。**Pi
 
 ### acp_delegate — 干净上下文委派
 
-把一个自包含的任务交给一个运行在干净上下文中的新 pi 进程。五个内置角色,各自有定制的工具白名单和系统提示:
+把一个自包含的任务交给一个运行在干净上下文中的新 pi 进程。五个内置角色,各自有系统提示和**软工具护栏**:
 
 | 角色 | 工具 | 适用场景 |
 |------|------|----------|
-| `reviewer` | read, bash | 只读代码审查(bug、风险、file:line) |
-| `researcher` | read, bash | 只读代码库调研 |
+| `reviewer` | read, bash, grep, find, ls + ACP | 只读代码审查(bug、风险、file:line) |
+| `researcher` | read, bash, grep, find, ls + ACP | 只读代码库调研 |
 | `worker` | read, edit, write, bash | 修改代码 |
-| `planner` | read, bash | 分析 + 提出分步计划 |
-| `oracle` | read, bash | 回答问题 / 建议 |
+| `planner` | read, bash, grep, find, ls + ACP | 分析 + 提出分步计划 |
+| `oracle` | read, bash, grep, find, ls + ACP | 回答问题 / 建议 |
 
-委派的完整结果保存到文件(`/tmp/acp-delegate/<runId>.out`);工具结果和注入通知只携带**任务标题 + 文件路径**(无预览)— 需要细节时用 `read` 读取。这让父上下文保持精简。
+只读角色(reviewer、researcher、planner、oracle)获得受限工具白名单(`read, bash, grep, find, ls`)+ ACP 上下文工具(`compress, decompress, search_context, acp_status`),以便管理自己的上下文。这能防止意外修改文件,但 `bash` 可绕过 - **这是护栏,不是安全边界**。
+
+Worker 运行在 Pi 的完整默认工具集上 - 不应用 `--tools` 白名单,因此任何已加载的扩展或自定义工具(如 ACP、LSP、MCP)保持可用。这确保主任务委派能力完整。上表中的 `read, edit, write, bash` 仅反映核心工具。
+
+委派的完整结果保存到文件(`/tmp/acp-delegate/<runId>.out`);工具结果和注入通知只携带**任务标题 + 文件路径**(无预览)- 需要细节时用 `read` 读取。这让父上下文保持精简。
 
 - **交互(TUI)与 RPC 模式**:`async:true`(默认)在后台运行子进程;完成时一条简短通知注入到聊天框。
 - **Print / JSON 模式**(`pi -p`、SDK):`async:true` 自动降级为**同步** — 结果在同一轮作为工具结果返回(父进程一轮后即退出,后台注入会丢失)。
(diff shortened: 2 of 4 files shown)


ranxianglei:
Thanks for this PR — the intent (making the delegated child process respect each role's tool whitelist) makes sense. But after digging into it, I think there are some deeper issues worth discussing before merging, because I'm not sure this actually achieves the access control it implies.

**1. The whitelist doesn't truly enforce read-only.**

Every read-only role (reviewer / researcher / planner / oracle) still gets `bash`. `bash` is a strict superset of `edit` / `write`:

```
echo "..." > file          # write
sed -i 's/old/new/' file   # edit
rm file                    # delete — even beyond edit/write
```

So "the reviewer can't edit/write" only holds for the *direct tool call*. A single shell command bypasses it completely. From an access-control standpoint the boundary is leaky.

**2. The dilemma this creates.**

The only way to make read-only *real* is to drop `bash` from those roles (keep just `read`, maybe `grep` / `find` / `ls`). But then a reviewer often *can't do its job* — searching the codebase, running tests/builds to validate the review, etc. all need the shell. So we're stuck between "not enforceable" and "can't work."

**3. Tool-list configuration gets complicated fast.**

As soon as we start curating per-role tool lists, a lot falls into scope:
- Our own context-management tools (`compress` / `decompress` / `search_context` / `acp_status`) — arguably the delegated child should also get these, otherwise it can't manage its own context under ACP.
- Third-party tools the user has installed and wants the child to use.
- Today the whitelist is hardcoded per role; supporting the above means making it *configurable*, and getting that configuration right is non-trivial.

**4. Precedent: `subagents`.**

I looked into `subagents` (another delegate-style extension) for comparison and hit the same wall: there's no clean way to make a delegated agent inherit the ACP compression pipeline. Either we inject/override the tool set, or we patch it at install time — neither is clean. `subagents` effectively can't support our context compression out of the box.

**5. Delegating the "main" task.**

There are also scenarios where a user delegates the *primary* task (not just a read-only review) to a child agent. If we now require per-role tool configuration — or push that config burden onto the user — the setup becomes heavy, which hurts the common case.

**My takeaway.**

I suspect this is less a flaw in this PR and more a fundamental tension in how Pi lets extensions sandbox/shape a child process: `--tools` is a *capability hint*, not a *capability boundary* (because `bash` exists), and there's no first-class mechanism in Pi to say "this role is truly read-only" or "this child should inherit these extension tools."

So I'd like to **open this up for discussion** rather than merge as-is. Some possible directions:

- **Accept as a soft guardrail:** merge this PR but clearly document that it only prevents *accidental/direct* edit/write and advertises the right toolset — it is *not* access control.
- **Push further:** define read-only roles without `bash` (+ grep/find/ls), accept the reduced capability, and work out the config story for ACP / third-party tools.
- **Reconsider the isolation unit:** whether delegate-style child processes are the right boundary at all vs. something Pi should provide natively.

Would love @ranxianglei's take too. Happy to help prototype whichever direction we land on.

Zeno2019:
Good points - I agree with the framing. Direction A works, with one addition from your own point 3.

**1. Yes, it's a soft guardrail, not a boundary.**

bash bypasses any tool allowlist - GuardFall (Adversa AI, 2026-06) bypassed 10 of 11 open-source agents this way. Hard isolation needs a system-level boundary, which is Pi's job, not BCP's.

What this PR actually does:
- stops read-only roles from *accidentally* editing files when prompt-following fails
- makes the README's "tailored tool whitelist" claim true

Guardrails prevent accidents, not attacks.

**2. Your point 3 has a built-in fix: auto-append the ACP tools.**

You said:

> "arguably the delegated child should also get these, otherwise it can't manage its own context under ACP."

Agreed - and it's ~4 lines, no config needed. pi-subagents does exactly this: `getSubagentToolAllowlist()` (src/tools/policy.ts) appends its protocol tools (`caller_ping`, `subagent_done`) to the child's allowlist. Same pattern for BCP:

```ts
const ACP_TOOLS = ["compress", "decompress", "search_context", "acp_status"];
if (agentDef?.tools) {
  const merged = [...new Set([...agentDef.tools.split(",").map(s => s.trim()), ...ACP_TOOLS])];
  cliArgs.push("--tools", merged.join(","));
}
```

Verified end-to-end (Pi v0.84.x):

```
pi -p --no-session --tools read,bash "call acp_status"
-> TOOL_NOT_AVAILABLE

pi -p --no-session --tools read,bash,compress,decompress,search_context,acp_status "call acp_status"
-> works, returns the ACP context breakdown
```

Why the append alone is sufficient: extension tools go through the same allowlist filter as built-ins (`agent-session.js:1954-1956`), and allowlisted registered tools auto-activate when `--tools` is present (`agent-session.js:1996-2000`). No settings.json, no env vars, no new config surface.

**3. Scope: read-only roles only. Worker stays unrestricted.**

worker's declared tools equal Pi's default active set, so `--tools` would only strip extension tools (`acp_delegate`, third-party tools) from the one role that handles primary-task delegation - your point 5 scenario. So `--tools` applies to reviewer/researcher/planner/oracle; worker opts out.

**4. README wording.**

"tailored tool whitelist" -> "tailored toolset per role (soft guardrail - prevents accidental edit/write; bash can bypass; not access control)".

**5. Deliberately not doing:**

- denylist / spawning controls - the allowlist already blocks nested delegation for read-only roles
- per-role user configuration - the complexity you flagged in point 3
- B/C - your call; happy to help prototype if you pick one

If this sounds right, I'll update the PR: ACP append + worker exemption + README rewording.

ranxianglei:
Thanks — great response. Direction A works for me; go ahead and update the PR with the three changes. Two things to nail down so the implementation matches the intent, plus one optional nit.

**1. Worker exemption needs a concrete mechanism.**

Right now all five roles carry a `tools` field (`delegate-tool.ts:30-60`), so the current `if (agentDef?.tools)` gate applies `--tools` to worker too — which is exactly the case we want to avoid (your point 3 / my point 5). "Worker opts out" has to be encoded explicitly. Cleanest option: add a flag to the role definition and gate on that instead of on the presence of `tools`:

```ts
interface AgentDef {
  prompt: string;
  tools: string;
  restricted?: boolean;   // true → emit --tools; absent → full default toolset
}
```

Set `restricted: true` on reviewer / researcher / planner / oracle; leave it off worker. Then `buildChildArgs` emits `--tools` only when `agentDef.restricted === true`, and does the ACP-tools append inside that branch. That keeps worker on Pi's full default toolset (which already includes the ACP tools, since no `--tools` filter means all registered tools stay active).

**2. Small correction on the rationale.**

"Worker's declared tools equal Pi's default active set" isn't quite accurate — Pi's default active set is larger than `read,edit,write,bash` (it also includes `glob`, `grep`, `find`, `ls`, `lsp_*`, etc.). So applying `--tools read,edit,write,bash` to worker would strip those builtins too, not just extension tools. The worker exemption matters a bit more than that framing suggests — good thing we're doing it. Not a blocker, just want the rationale airtight.

**3. ACP-tools append — agreed, ship it.**

The `getSubagentToolAllowlist()` precedent is the right model and the v0.84.x end-to-end check is convincing. Just make sure the append runs inside the `restricted` branch (worker doesn't need it — it gets everything via the default set). Your snippet is already correct on that once it's gated on the new flag.

**Nit (optional):** the read-only roles are currently `read,bash`; `grep` / `glob` / `find` / `ls` aren't listed (bash covers them as shell commands, so it's functional). If you'd like the whitelist to be more ergonomic/honest, listing them explicitly wouldn't hurt — purely optional, your call.

**Update checklist:**
- [ ] `AgentDef.restricted?: boolean`; set on the 4 read-only roles
- [ ] `buildChildArgs`: emit `--tools` only when `restricted`, with ACP tools appended
- [ ] README rewording (your wording is good)
- [ ] *(optional)* explicit `grep,glob,find,ls` in read-only tools

Looking forward to the update.

Zeno2019:
> Thanks — great response. Direction A works for me; go ahead and update the PR with the three changes. Two things to nail down so the implementation matches the intent, plus one optional nit.
> 
> **1. Worker exemption needs a concrete mechanism.**
> 
> Right now all five roles carry a `tools` field (`delegate-tool.ts:30-60`), so the current `if (agentDef?.tools)` gate applies `--tools` to worker too — which is exactly the case we want to avoid (your point 3 / my point 5). "Worker opts out" has to be encoded explicitly. Cleanest option: add a flag to the role definition and gate on that instead of on the presence of `tools`:
> 
> ```ts
> interface AgentDef {
>   prompt: string;
>   tools: string;
>   restricted?: boolean;   // true → emit --tools; absent → full default toolset
> }
> ```
> 
> Set `restricted: true` on reviewer / researcher / planner / oracle; leave it off worker. Then `buildChildArgs` emits `--tools` only when `agentDef.restricted === true`, and does the ACP-tools append inside that branch. That keeps worker on Pi's full default toolset (which already includes the ACP tools, since no `--tools` filter means all registered tools stay active).
> 
> **2. Small correction on the rationale.**
> 
> "Worker's declared tools equal Pi's default active set" isn't quite accurate — Pi's default active set is larger than `read,edit,write,bash` (it also includes `glob`, `grep`, `find`, `ls`, `lsp_*`, etc.). So applying `--tools read,edit,write,bash` to worker would strip those builtins too, not just extension tools. The worker exemption matters a bit more than that framing suggests — good thing we're doing it. Not a blocker, just want the rationale airtight.
> 
> **3. ACP-tools append — agreed, ship it.**
> 
> The `getSubagentToolAllowlist()` precedent is the right model and the v0.84.x end-to-end check is convincing. Just make sure the append runs inside the `restricted` branch (worker doesn't need it — it gets everything via the default set). Your snippet is already correct on that once it's gated on the new flag.
> 
> **Nit (optional):** the read-only roles are currently `read,bash`; `grep` / `glob` / `find` / `ls` aren't listed (bash covers them as shell commands, so it's functional). If you'd like the whitelist to be more ergonomic/honest, listing them explicitly wouldn't hurt — purely optional, your call.
> 
> **Update checklist:**
> 
> * [ ]  `AgentDef.restricted?: boolean`; set on the 4 read-only roles
> * [ ]  `buildChildArgs`: emit `--tools` only when `restricted`, with ACP tools appended
> * [ ]  README rewording (your wording is good)
> * [ ]  _(optional)_ explicit `grep,glob,find,ls` in read-only tools
> 
> Looking forward to the update.

Zeno2019:
Thanks - I updated the PR along Direction A:

- Added a `restricted` flag for reviewer, researcher, planner, and oracle.
- Restricted roles now receive `read`, `bash`, `grep`, `find`, and `ls`, plus the ACP context tools.
- Worker remains unrestricted, so loaded extension and custom tools are still available.
- Updated the README to clarify that this is a guardrail against accidental edits, not a security boundary. `bash` can still modify files.
- Added tests for the restricted roles, worker, and unknown agents.

I left out `glob` because it is not registered as a Pi agent tool. `grep`, `find`, and `ls` are registered core tools, while `lsp_*` tools come from extensions.

Tests, typecheck, and build are passing.

ranxianglei (approved):
LGTM — verified locally against master (5175def): diff applies cleanly, `typecheck` clean, **58/58 tests pass** (48 base + 10 new), `build` success.

Confirmed the implementation matches Direction A:
- `restricted?: boolean` on `AgentDef`; set on the 4 read-only roles, absent on worker.
- `buildChildArgs` gates `--tools` on `agentDef?.restricted` (not on the presence of `tools`), so worker + unknown agents stay on Pi's full default toolset.
- ACP tools (`compress`/`decompress`/`search_context`/`acp_status`) auto-appended for restricted roles via a de-duping Set merge.
- README + README.zh-CN both reworded to make the soft-guardrail framing explicit.

Also verified the Pi tool-name choices against pi-stable source: Pi's core tools are exactly `read, edit, write, bash, grep, find, ls` — `glob` is not a Pi tool (correctly omitted), and `grep`/`find`/`ls` are real core tools (correctly included). Approving.

PR #91, opened 2026-08-09
feat(delegate): watchdog + live activity file & reply stream (json event mode)

Tyan66666:
## Background

Ref #89 — two asks: (1) see what a delegate is doing **while it runs**; (2) auto-terminate delegates that **hang** (provider unresponsive → child never exits → run stuck in "running" forever).

## Watchdog (was PR #90, folded in)

Async delegates are force-finished when: no output for 5m (idle — the main defense, since a hung child holds its stdout fd open so EOF never fires), 10s after output ends, or a 30m hard limit. Sequence: SIGTERM → 10s grace → SIGKILL. The result reflects whatever was produced, with a `(timed out: <reason>)` header.

Logic extracted to `src/delegate-watchdog.ts` (`attachWatchdogs(child, hooks, opts)`, injectable timeouts) — 7 unit tests in `tests/watchdog.test.ts` (fake child + short timers): idle SIGTERM, SIGKILL escalation, poke reset, EOF grace, hard limit, dispose, settled guard.

## Live activity file + reply stream

Async delegates now spawn with `--mode json` instead of `-p`. The host parses the newline-delimited event stream into **two live files**:

| File | Content | When you see it |
|---|---|---|
| `.out` | assistant reply token stream (`text_delta` accumulations, `text_end` authoritative) | delivered as the result file on finish |
| `.activity` | human-readable tool activity — `[tool] bash <cmd>`, live output chunks, `[done] bash` | streamed in real time, readable anytime |

Key points:

- **Real-time bash output**: `tool_execution_update.partialResult` is written as it arrives — verified `tick-1`/`tick-2` visible mid-run while the command still sleeps.
- **Snapshot dedup**: `partialResult` is an accumulated snapshot (not a delta) — only the newly-appended portion is written per update.
- **`showThinking` option** (default `false`): thinking deltas merged per segment into one `[thinking] <full segment>` line instead of token fragments.
- **Provider retry visibility**: `auto_retry_start` / `auto_retry_end` events (transient errors: overloaded / rate limit / 5xx) are formatted as `[retry] attempt 1/3, backoff 2000ms — 529 Overloaded` and `[retry] attempt 2 succeeded` — the original #89 scenario (provider unresponsive → delegate retrying unseen) is now observable.
- **Spawn returns only the `.activity` path**; the `.out` path arrives with the result notification.
- **Sync delegates keep `-p`** — no streaming, no behavior change; the async auto-downgrade for print/json hosts stays safe (buildChildArgs computes the same `isAsync`).

## Tests

20 new unit tests (7 watchdog + 13 events) using real event samples captured from `pi --mode json`. 103 total, typecheck + build green. E2E verified on a local install: reply extraction, mid-run activity visibility, snapshot dedup, merged thinking, `.out` delivery, watchdog idle kill (exit 143 + `(no output)` result).

⚠️ This PR **replaces** PR #90 (closed). Watchdog and activity-stream land as one unit.

### DIFF
--- a/src/delegate-events.ts
+++ b/src/delegate-events.ts
@@ -1,0 +1,174 @@
+// Parsing helpers for `pi --mode json` event streams. The child delegate runs
+// in JSON Event Stream Mode, so its stdout is newline-delimited JSON events
+// instead of the final reply text. These pure functions map raw lines to
+// structured events and format them for the activity file.
+
+export interface ToolStartEvent {
+  kind: "tool-start";
+  toolName: string;
+  argsText: string;
+}
+
+export interface ToolUpdateEvent {
+  kind: "tool-update";
+  toolCallId: string;
+  text: string;
+}
+
+export interface ToolEndEvent {
+  kind: "tool-end";
+  toolName: string;
+  isError: boolean;
+}
+
+export interface ReplyDeltaEvent {
+  kind: "reply-delta";
+  delta: string;
+}
+
+export interface ReplyCompleteEvent {
+  kind: "reply-complete";
+  content: string;
+}
+
+export interface ThinkingDeltaEvent {
+  kind: "thinking-delta";
+  delta: string;
+}
+
+export type ParsedEvent =
+  | ToolStartEvent
+  | ToolUpdateEvent
+  | ToolEndEvent
+  | ReplyDeltaEvent
+  | ReplyCompleteEvent
+  | ThinkingDeltaEvent
+  | ThinkingEndEvent;
+
+export interface ThinkingEndEvent {
+  kind: "thinking-end";
+}
+
+/**
+ * Accumulates thinking_delta tokens and emits one human-readable line per
+ * thinking segment (a segment ends at thinking_end / text_start). Nothing is
+ * emitted when showThinking is off.
+ */
+export class ThinkingCollector {
+  private buf = "";
+
+  constructor(private readonly showThinking: boolean) {}
+
+  push(delta: string): void {
+    this.buf += delta;
+  }
+
+  /** Return the segment line to write ("" when empty or disabled), resetting. */
+  flush(): string {
+    const text = this.buf.trim();
+    this.buf = "";
+    if (!this.showThinking || !text) return "";
+    return `[thinking] ${text}\n`;
+  }
+}
+/** Parse one newline-delimited JSON line; null for non-JSON or irrelevant events. */
+export function parseEventLine(line: string): ParsedEvent | null {
+  let ev: unknown;
+  try {
+    ev = JSON.parse(line);
+  } catch {
+    return null;
+  }
+  if (typeof ev !== "object" || ev === null) return null;
+  const e = ev as Record<string, unknown>;
+  if (e.type === "message_update") {
+    const am = e.assistantMessageEvent;
+    if (typeof am !== "object" || am === null) return null;
+    const msg = am as Record<string, unknown>;
+    switch (msg.type) {
+      case "text_delta":
+        return { kind: "reply-delta", delta: String(msg.delta ?? "") };
+      case "text_end":
+        return { kind: "reply-complete", content: String(msg.content ?? "") };
+      case "thinking_delta":
+        return { kind: "thinking-delta", delta: String(msg.delta ?? "") };
+      case "thinking_end":
+        return { kind: "thinking-end" };
+      default:
+        return null;
+    }
+  }
+  if (e.type === "tool_execution_start") {
+    return {
+      kind: "tool-start",
+      toolName: String(e.toolName ?? ""),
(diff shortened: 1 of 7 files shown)


ranxianglei:
let me check oh-my-pi 是否兼容

ranxianglei:
Thanks for this — the watchdog half (idle/EOF/hard-limit kill) is exactly right and targets the real root cause (a hung child keeps its stdout fd open so `close` never fires). One compatibility concern with the streaming half before this lands:

**`--mode json` is pi-only and breaks under omp (oh-my-pi).**

This adapter is also expected to run under [`can1357/oh-my-pi`](https://github.com/can1357/oh-my-pi) (we already feature-detect it — see `src/runtime.ts`, `buildContextEntries()` vs `getBranch()`). omp is **not** a thin pi fork: it is a heavily-rewritten pi-mono derivative (~80k lines of Rust core, Bun runtime, its own `task` subagent system). omp does **not** support `--mode json` — its documented entry points are only:

- interactive (default TUI)
- `-p` (one-shot)
- `--mode rpc`
- `--mode rpc-ui`
- `omp acp`

There is no `json` mode. So with the current `buildChildArgs` (async ⇒ `--mode json` unconditionally), an async delegate spawned under omp will fail to start or emit an unparseable/unmatched event stream. And even if omp silently accepted the flag, its event schema (Rust core) would not match the `text_delta` / `thinking_delta` / `tool_execution_update` / `auto_retry_*` events that `delegate-events.ts` parses.

**Suggested fix:** keep the watchdog host-agnostic (it already is), but gate the `--mode json` streaming behind host detection — under omp fall back to the existing `-p` spawn (loses live streaming but keeps the hang fix and stays functional). We already have the pi-vs-omp probe in `runtime.ts`, so it is a small change, e.g.:

```ts
const useJsonStream = typeof ctx.sessionManager?.buildContextEntries === "function"; // pi; omp only has getBranch
cliArgs = isAsync && useJsonStream
  ? ["--mode", "json", "--no-session", "--append-system-prompt", promptFile]
  : ["-p", "--no-session", "--append-system-prompt", promptFile];
```

This makes async delegates work on both hosts. Happy to help test either path if useful.

Tyan66666:
## omp compatibility fix (thanks for the feedback!)

Implemented the host-detection gate you suggested and went one step further: the stdout consumer had to change too, otherwise a `-p` fallback would silently drop all output (the async path only parsed json event lines).

**What changed** (`c4bbba2`):

- **`src/runtime.ts`** — new `isPiHost(sm)`: true only when `sessionManager.buildContextEntries` exists (same feature-detect as `readContextEntries`); omp only exposes `getBranch` → false.
- **`src/delegate-tool.ts`**
  - `buildChildArgs` now returns `{ cliArgs, tmpDir, isAsync, useJsonStream }`; `useJsonStream = isAsync && isPiHost(...)`. omp async delegates spawn `-p` instead of `--mode json`.
  - stdout handler branches: json mode parses event lines as before; omp mode writes stdout straight through as the reply (no line buffering, so a trailing partial chunk is kept — no EOF flush needed).
  - No `.activity` file is created on omp (`activityStream` is nullable; all writes, `endStream` and `destroy` tolerate null), and the spawn message points at the `.out` reply file instead of the activity file.
  - `debug.event("delegate-spawn")` now logs `useJsonStream` too.
- **`tests/delegate-tool.test.ts`** — `mockCtx` gains a pi/omp host switch + 3 new cases: pi async → `--mode json`, omp async → `-p`, pi sync → `-p`.

**Verification**: `npm run typecheck` ✅ · `npm test` ✅ (110/110) · `npm run build` ✅

The watchdog (idle / EOF-grace / hard-limit) is host-agnostic and still applies unchanged on omp.

ranxianglei:
Reviewed `c4bbba2` — this is the right fix, thanks. A few things I checked:

- `isPiHost()` reuses the exact same `buildContextEntries` detection as `readContextEntries`, so pi/omp discrimination stays consistent in one place. Good.
- `useJsonStream = isAsync && isPiHost(...)` — pi async → `--mode json`, omp async → `-p`, sync → `-p`. Correct matrix.
- The stdout-consumer branch is the important catch: under omp `-p` the plain reply goes to stdout, so streaming it straight into the `.out` reply file (no line buffering, trailing partial chunk kept) is exactly right — a json-only parser would have silently dropped everything. 
- `watchdog.poke()` sits above the `if (useJsonStream)` branch, so the idle/EOF-grace/hard-limit watchdog still fires on omp too. The hang fix is genuinely host-agnostic now.
- `activityStream` nullable + all `?.write` / `?.destroy` / null-tolerant `endStream` — no stray writes on the omp path.

The one inherent omp limitation (async delegates there only surface the final reply to `.out`, no live tool activity) is unavoidable without a json event mode and is clearly communicated in the spawn message. LGTM from a compatibility standpoint.

PR #181, opened 2026-08-20
fix(compress): reject tier-3-only rewrite ranges (infinite re-condense loop)

ranxianglei:
Closes #3 (fix portion).\n\n## Root cause\nacp-kernel `applySingleRange` computes `outputTier = min(3, targetTier + 1)`. Compressing an already-tier-3 block range (`compress(bN..bN)`) therefore always "succeeds" (blocksCreated=1, no error, no warning), consuming the old T3 block and producing a new T3 block. Because every call succeeds, the 3-attempt retry cap (which only counts failures) never triggers and the nudge baseline resets each time — an infinitely repeatable rewrite of the same summary.\n\nIndependently reproduced on pinned acp-kernel 0.0.28: b3→b4→b5→b6→b7→b8, each call created=1, errors=0.\n\n## Fix (adapter side)\n- `src/compress-tool.ts`: detect a degenerate outcome — new block tier=3, `directMessageIds` empty, only T3 blocks consumed — and treat it as a tier-3-only rewrite: roll back to the pre-apply state and throw (isError path), guiding the model to `search_context`/`decompress` or a range containing uncompressed messages. The existing per-turn retry cap scopes follow-up retries.\n- `tests/t3-rewrite-guard.test.ts`: 2 regression tests (rewrite rejected + state rolled back; legitimate T1→T2→T3 distillation unaffected).\n\nVerified: `tsc --noEmit` clean, 384 tests pass (incl. 2 new), `npm run build` OK.\n\nComplementary to #180 (retry-cap bypass via 0-block panels) — neither fix covers the other's hole; both are needed.\n\nKernel-side root cure (follow-up, requires acp-kernel release first per AGENTS.md): reject ranges where `outputTier <= targetTier` and `directMessageIds` is empty in `applySingleRange`.\n\n(Branch authored in session for issue #3; PR opened from audit in issue #5. Merge is human-only.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,13 +5,13 @@
 	"requires": true,
 	"packages": {
 		"": {
-		"name": "billion-context-pi",
-		"version": "0.1.42",
-		"license": "MIT",
+			"name": "billion-context-pi",
+			"version": "0.1.42",
+			"license": "MIT",
 			"devDependencies": {
 				"@earendil-works/pi-coding-agent": "0.83.0",
 				"@types/node": "^26.1.2",
-				"acp-kernel": "0.0.28",
+				"acp-kernel": "0.0.29",
 				"billion-context-kit": "0.2.0",
 				"tsup": "^8.5.1",
 				"tsx": "^4.23.1",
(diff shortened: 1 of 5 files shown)


ranxianglei:
[bot] Added 1a3145b — acp-kernel 0.0.28 → 0.0.29 (exact pin, lockfile synced).

Scope of 0.0.29: search-only (Intl.Segmenter CJK tokenizer, kernel PR #87); compress/boundaries code untouched. Validation on 0.0.29: tsc clean, 384/384 tests, build OK.

⚠️ Correction to my earlier note in #3: the multi-block-distill anchor fix (kernel commit 79340d3, "snap consumed anchors", kernel #32) is **NOT part of 0.0.29** — it is unreleased on kernel master. It needs a kernel 0.0.30 release first; then this repo bumps again per the AGENTS.md publish order (kernel → verify on npm → bump here). Kernel-side follow-ups for 0.0.30: (1) anchor snap (79340d3, already on kernel master), (2) reject pure-T3 rewrite in `applySingleRange` (adapter-side guard here already covers it).

PR #196, opened 2026-08-22
feat: salvage malformed compress content strings (omp#121)

ranxianglei:
## Why

Weak/local models (vLLM qwen etc.) emit compress tool arguments that fail strict JSON.parse ~50% of the time — truncated output caps, raw newlines inside summary strings, trailing commas, or plain prose (see billion-context-omp#121). The ACP host extension's `normalizeRanges` threw `Invalid content: not valid JSON` on the first strict-parse failure and relied entirely on the retry nudge — but weak models fail the *retry* the same way ~50% of the time, so compress became effectively unusable with local models.

## What

Before throwing, run the kernel's `salvageParseRanges` ladder (fences / comma+newline repairs / truncated-array prefix / field-regex): recoverable entries proceed as normal ranges (logWarn records the layer); truly unparseable input still throws, so the `isError:true` → retry-nudge contract is unchanged.

Bumps acp-kernel to ^0.0.34 (introduces `salvageParseRanges` + `extractRanges` double-encode normalization).

## Tests

New truncated-content salvage case in `tests/compress-retry.test.ts` (first complete entry recovered from a blob truncated mid-second-entry). Full suite 410/410, typecheck clean.

Companion PRs: acp-kernel#109 (parser), billion-context (proxy), billion-context-omp (replay).

--- DEPENDENCY NOTE (pre-publish) ---
> ⚠️ **Do not merge yet**: `package.json` currently pins `acp-kernel` to a local `file:` tarball so the diff is reviewable now. Depends on acp-kernel PR #109 landing and npm publish of **0.0.34** (0.0.33's publish failed with npm 404 — expired NPM_TOKEN). Before merge this PR will be updated to `"acp-kernel": "^0.0.34"` from from the registry and the full suite re-run.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
 			"devDependencies": {
 				"@earendil-works/pi-coding-agent": "0.83.0",
 				"@types/node": "^26.1.2",
-				"acp-kernel": "0.0.32",
+				"acp-kernel": "file:../acp-kernel/acp-kernel-0.0.33.tgz",
 				"billion-context-kit": "0.2.0",
 				"tsup": "^8.5.1",
 				"tsx": "^4.23.1",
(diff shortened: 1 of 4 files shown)


ranxianglei:
关闭:已被内核 + master 侧修复吸收(superseded)。\n\n- 内核 0.0.81 `parseCompressArgs` 自带完整宽容解析(fenced/尾逗号/裸换行/double-stringified/截断前缀 salvage);\n- master `src/compress-tool.ts` 走内核解析,另有 #481 的 `arrayWrapRepair`(裸对象串重包,#480)与 `tailRepair`(缺 `}` 修补),覆盖面超过本 PR;\n- 本 PR 引用的 `salvageParseRanges` 导出在 0.0.81 已不存在,分支已无法编译。

PR #215, opened 2026-08-23
fix(overflow): arm self-heal on ambiguous no-body 4xx errors

ranxianglei:
# fix(overflow): arm self-heal on ambiguous no-body 4xx errors

> Model: **GLM-5.3** (zhipuai-lb, via pi coding agent) — per AGENTS.md rule 2.
> Branch: `2026-08-23_overflow-nobody-arm` (based on `origin/master` @ 5a834ef) · Commit: `c355c4e`

## Problem

**Real incident (2026-08-23):** an ACP-managed pi session dead-looped. A 50,358-char bash `toolResult` (~31,475 tokens, ~24% of the effective 131,072-token window) pushed the request past sglang's `input + max_tokens ≤ 262,144` hard cap. Every subsequent request returned `400 status code (no body)` forever:

- The model never got a successful turn, so it could never call `compress` — the ACP loop never shrank the context.
- The user's "继续" just resent the same oversized context.
- The extension's overflow self-heal never fired: it only arms on `OVERFLOW_MARKER` text markers (`maximum context length is N`, `prompt is too long`, …), and a bodyless 4xx carries none of them.
- pi itself classifies this text as overflow (pi-ai `OVERFLOW_PATTERNS` ends with `/^4(?:00|13)\s*(?:status code)?\s*\(no body\)/i`), but the extension-side marker set deliberately never matched it — for good reason: OpenAI-compatible providers return the exact same no-body 400 for **non-overflow** reasons (invalid model, malformed request; see `src/messages.ts:80-82`), so treating it as an unconditional overflow risks false positives.

## Fix

Treat pi's bodyless `4xx ... (no body)` error text as a **possible** overflow signal, and arm the emergency self-heal (forced ≥95% usage → kernel emergency nudge + tool-result truncate) only when at least one false-positive guard holds:

1. **Ratio guard:** ACP's current sent-view token estimate ≥ 50% (`NO_BODY_ARM_RATIO`) of the effective limit — the same pct basis already logged per turn (`tokenCount / config.modelContextLimit`, recorded per request in the `context` event, before the armed boost so the guard sees the real estimate).
2. **Consecutive guard:** it is the ≥2nd consecutive no-body 4xx since the last successful assistant turn. Count resets on a successful assistant turn and at session start (per-session `OverflowEpisode`), recovering the dead-loop even when the estimate under-reports (the incident ran at ~24%: the sent-view estimate cannot see the provider's `input+max_tokens` accounting).

When the body has no parseable limit number (always, for a bodyless error), no window is learned — the armed emergency uses the already-resolved effective limit, so arming never fails for lack of a parsed number.

The classic text-marker arm path (`inspectOverflowMessage` → learn window + arm) is **unchanged**.

### Files changed

| File | Change |
|---|---|
| `src/overflow-selfheal.ts` | `NO_BODY_4XX_MARKER` + `isNoBody4xxError()` (mirrors pi-ai's anchored regex); `NO_BODY_ARM_RATIO = 0.5`; `OverflowEpisode.noteSentView/noteSuccess/onNoBody4xx()` (returns `{arm, consecutive, ratio}`); `reset()` clears the new state. Guard state is runtime/in-memory like the `armed` flag — not persisted to acp.json (justified in comment: a resumed session re-establishes both within one turn; a persisted count would arm against a fresh session whose first request may succeed). |
| `src/index.ts` | `wireContextTransform`: record `noteSentView(tokenCount, config.modelContextLimit)` before the armed boost. `wireOverflowSelfHeal` (`message_end`): non-error assistant turns call `noteSuccess()`; error turns run the classic marker path unchanged, else the no-body path with the two guards; arms + logs `no-body-arm` (consecutive, ratio) + UI notify. |
| `tests/overflow-selfheal.test.ts` | 10 new tests (21 → 31 in file). |
| `CHANGELOG.md` | Unreleased entry. |

## Tests

`npm run typecheck` ✅ · full suite `npm test` ✅ **422 pass / 0 fail** (was 412; +10):

- `isNoBody4xxError` matches `400/413 status code (no body)`, `400 (no body)`, `413(no body)`; rejects `429/500/404 (no body)`, quota text, empty, and prefixed text (anchored, like pi-ai).
- no-body at low estimate (~24%, incident scale) + first occurrence → **NO arm**.
- no-body at ≥50% estimate → arm on first occurrence (incl. exact-threshold `>=` case).
- first no-body at low estimate, then a second consecutive no-body → **arm** (dead-loop recovery).
- classic `maximum context length is 262144 ...` → still arms via the text-marker path, window parsed (regression).
- successful turn resets the consecutive count; `OverflowEpisode.reset()` clears sent-view + count; fresh episode (no recorded sent view) does not arm on the first hit.

## Notes

- Complements **acp-kernel PR #133** (kernel-side counterpart of the no-body overflow handling).
- Relates to open issue **#204** ("90% 不截断导致上下文爆掉") — same failure family: the context grew past what the provider accepts and no mechanism shrank it in time; this PR guarantees an emergency shrink path even when the provider reports the overflow as a bodyless 4xx.
- No behavior change for non-no-body errors; throttle-retry (`isThrottleError`) is unaffected (a no-body 4xx never matched its throttle phrases).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased (master, since v0.1.38)
+- **fix(overflow)**: 无 body 4xx 也触发溢出自愈 — pi 把 provider 空 body 的 4xx 原样透出为 `400/413 status code (no body)`（pi-ai 自身按 `/^4(?:00|13)\s*(?:status code)?\s*\(no body\)/i` 归类为 overflow），但 OVERFLOW_MARKER 不含该形态，扩展侧自愈永不武装 → 死循环无法恢复（2026-08-23 事故：50k 字符 bash toolResult 撞穿 sglang input+max_tokens 硬上限，此后每轮 400 无 body，模型永远等不到成功回合去 compress，用户「继续」只是原样重发超限上下文）。现将其作为**疑似溢出**信号：仅当发送视图估算 ≥ 有效上限 50%（与 turn 日志 pct 同口径）或自上次成功 assistant 回合起连续 ≥2 次无 body 4xx 时才武装紧急压缩；成功回合/新会话清零计数（runtime 内存态，与 armed 同生命周期，不入 acp.json）。无 body 解析不出窗口数 → 不学习窗口，紧急压缩直接用已解析的有效上限。经典文本标记路径不变 (relates #204)
 - **fix(guardrail): `toolOutputMaxBytes` 未配置时文档默认值 200000 现在实际生效** — 原接线 `if (max !== undefined && max > 0)` 把「未配置」当成「禁用」，内置 200KB 天花板永远不可达（`capToolOutput` 内部的回退到不了）；pi 只内置 cap bash/read/grep，其余工具可无限注入 context，与 CONFIGURATION.md 承诺的 ACTIVE 默认不符。改为接线层 `?? DEFAULT_TOOL_OUTPUT_MAX_BYTES` 回退，`0`/负数禁用语义不变 (#210)
 
 - **fix(compress): 接受 JSON 字符串形式的 `content` 参数** — 非严格工具 provider（vLLM openai-completions，`supportsStrictTools:false`）会把嵌套数组参数字符串化，pi 的 typebox 校验直接拒掉（`content.0: must be object`）。实测会话 01a00a38 全部唯一一次 compress 调用即死于此，3 小时会话零压缩。schema 改为 `Type.Union([Array, String])`，字符串自动 `JSON.parse` 并校验（错误信息引导模型传数组）
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,7 +32,7 @@ import {
 } from "./throttle-retry.js";
 import { defaultCountTokens } from "acp-kernel";
 import { formatSystemPromptForEvent, getSystemPromptText } from "./compat.js";
-import { inspectOverflowMessage, reserveOutputHeadroom, shouldReserveOutputHeadroom } from "./overflow-selfheal.js";
+import { inspectOverflowMessage, isNoBody4xxError, reserveOutputHeadroom, shouldReserveOutputHeadroom } from "./overflow-selfheal.js";
 
 type AgentMessage = SessionMessageEntry["message"];
 
@@ -195,6 +195,12 @@ function wireContextTransform(pi: ExtensionAPI, runtime: AcpRuntime): void {
       // below is fed the RAW sentTokens — its samples must stay on the
       // raw basis or density would chase its own calibration.
       let tokenCount = calibrateTokens(sentTokens, runtime.density.densityFor(modelId));
+      // Basis for the no-body-4xx overflow guard (wireOverflowSelfHeal): the
+      // sent-view estimate + effective limit of the request about to be sent —
+      // the same pct basis as the turn log below. Recorded BEFORE the armed
+      // boost so a guard consult after an error sees the real estimate, not
+      // the forced >=95% floor.
+      ov.noteSentView(tokenCount, config.modelContextLimit);
       // Self-heal (armed): after an overflow, force this turn's usage to >=95%
       // so the kernel's emergency nudge + tool-result truncate fire immediately,
       // even if the density-calibrated estimate under-reports the sent view.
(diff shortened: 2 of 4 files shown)


ranxianglei:
Review round 1 (agent reviewer — **verdict: ship**):

- **Regex matches pi's real surface form**: openai-completions.js:451 passes the openai SDK's verbatim `400 status code (no body)` with no prefix; pi-ai's own classifier uses the identical anchored regex. 401/403/429 can't match.
- **Healthy path unchanged**: success branch only adds `noteSuccess()`; nothing outbound; state in-memory only.
- **Emergency fires only during genuine recovery** (`armed` consumed once per arm; success resets; no healthy-path rewrite of old messages).

Minor:
- **False-positive surface is real but gated**: persistent non-overflow bodyless 400s are documented in-repo (GLM empty-assistant quirk, messages.ts:81; body-stripping relays). ≥2-consecutive arms → per-turn view truncation + nudge spam until the first success; originals survive in the session file. Accepted trade for the dead-loop it fixes.
- No wiring-level test pinning `noteSentView` *before* the armed boost (property exists only as a comment).
- **Merge order vs #214** (guaranteed src/index.ts conflict): place `ov.noteSentView(...)` **after** #214's anchored max-merge and **before** the armed boost — then the ratio guard sees the provider-anchored basis and arms on the *first* no-body error in an incident (anchored ~100% ≥ 50% immediately), instead of waiting for the second.

ranxianglei:
**分诊(队列清理 #484): KEEP-VALID。**

缺口在 master 上仍真实存在:`src/overflow-selfheal.ts:48-51` 的 `inspectOverflowMessage` 在 body 为空时直接返回 `{isOverflow:false}` —— 对 sglang/vLLM 空 body 的 400/413 无能为力,master 无任何 no-body 4xx 佐证逻辑(grep `NO_BODY_4XX|isNoBody4xxError` 零命中)。

设计上对误报的防备是周全的:仅当 N 次*连续* no-body 4xx(自上次成功 turn 起)+ 与记录的 sent-view 估算互证后才 arm,arm 后 boost 前仍查 guard。

**rebase 风险:重。** `wireContextTransform`(原 `noteSentView` 记录点)已被 #257/#455/#289 的 floors/calibration 重写,PR 所 patch 的 `calibrateTokens(sentTokens, runtime.density…)` 行已不存在;`wireOverflowSelfHeal` 移到了 `src/index.ts:823+`。两个钩子需按新代码手工重放。

ranxianglei:

✅ **Rebased onto current master and force-pushed** — branch `2026-08-23_overflow-nobody-arm` is now `55e549f` on top of `72d8771` (was `c355c4e` on stale base `5a834ef`). `mergeable: true`, no conflicts; GitHub CI 9/9 green (test ×4 OS/node matrix, e2e ×2, pr-validation, docker, build-artifact).

**How the two hooks were replayed** (manual re-application per triage):

1. **`wireContextTransform` sent-view recording** — the original patch point (`calibrateTokens(sentTokens, runtime.density…)`) no longer exists after the floors/calibration rewrite (#257/#455/#289). `ov.noteSentView(...)` is now called right after the view-recount block, recording a *guard basis* = `max(calibrated sent-view, provider-usage floor)` — deliberately **excluding the armed 95% floor**. That floor is an artifact of a prior arm; feeding it back would let the ratio guard re-arm unconditionally right after an emergency, defeating the corroboration design. The provider-usage floor stays in the basis because it's real size evidence. Placement is pre-armed-boost as originally intended (#214's anchored-max ordering guidance no longer applies — that PR closed unmerged).
2. **`wireOverflowSelfHeal` message_end** — relocated to its new position in `src/index.ts`; restructure applied verbatim from `c355c4e`: a successful turn resets the consecutive counter, and no-body 4xx arms only through `onNoBody4xx()` corroboration (ratio ≥ 0.5 against the recorded sent-view, or second consecutive hit).

**Verification on the rebased head:**
- typecheck clean
- full suite: failure set byte-identical to clean master (82 pre-existing local-environment failures, identical before/after this change, unrelated to it); `overflow-selfheal.test.ts` 36/36 pass including the 10 new tests
- production build OK (self-contained bundle)
- GitHub CI: 9/9 green

PR is ready for human merge.

**中文摘要**:已按分诊把 #215 的两个钩子手工重放到重写后的 master(sent-view 记录点改用排除 armed floor 的 guard basis,避免紧急压缩后无条件再 arm;message_end 逻辑原样重放到新位置),本地测试与干净 master 失败集一致、新增 10 个用例全过、GitHub CI 全绿,可以合并。

PR #228, opened 2026-08-25
fix(nudge): unify transient-injection budgets into per-turn injection ledger

ranxianglei:
## Summary

Implements the comprehensive governance plan from the #223 post-mortem audit (ework #18): one injection ledger replaces the whole family of ad-hoc retry/nudge budgets (#6 breaker, and the unmerged #7 / #9 / #194 / #217 branches).

### Root cause of the #223 amplification chain

pi rebuilds the outbound array on **every LLM call**, so any transient injection re-appends per fire unless budgeted. Every patch in the lineage measured a different quantity — Set dedup, distinct failed calls, fire budgets — none measured **injections**:

| mechanism | counted | escape |
|---|---|---|
| retry prompt (removed in #225) | distinct failed toolCallIds | model never retries → stuck at 1, re-injects forever (#223) |
| #6 nudge breaker | failed/no-op compress calls | `compress({content: []})` → `"No ranges provided."` isError:false → **neutral**, counter never moves → emergency nudge re-injects forever |
| throttle kick / delegate notify | (none) | synthetic user msgs rotate the turn key → reset budgets mid-loop |

### The fix

- **One ledger** (`runtime.noteInjection(turnKey, kind, budget)`): nudge = 1/genuine-turn, emergency = `MAX_EMERGENCY_NUDGES_PER_TURN` (3)/turn. Budgets count **injections** — ignoring, failing, no-op, or neutral compress responses cannot extend them. Only a genuine user-turn switch resets.
- **Genuine turn key**: `lastUserMessageId` (src/tokens.ts) now skips synthetic user messages (`[ACP:provider-throttle]` kicks, `[acp_delegate …]` notifications) — they no longer reset budgets.
- **Deleted ~150 LOC** of compress-outcome machinery whose only surviving customer was the emergency gate: `noteCompressOutcomes`, `compressOutcomeSeen`, `compressFailTurnKey/Count`, `compressRetryCappedFor`, `collectCompressOutcomes`, `isCompressSuccessText/isCompressNoopText`, and the separate `nudgeShownTurns` Set. The neutral escape dies with it.
- One-shot UI notice when the emergency budget exhausts ("nudge paused … emergency truncation still active"); kernel ≥95% truncation remains the mechanical backstop.

### Adversarial regression test

20 context fires alternating every response shape the old budgets could not reach (hard failure / no-op panel / neutral `"No ranges provided."` / plain silence) → **exactly 3 emergency injections**. Throttle-kick and delegate messages do not re-arm the budget; a genuine user message does. #223 regression (zero retry prompts ever) retained.

## Test plan

- [x] `npm run typecheck` ✓
- [x] `npm test` — 426/426 ✓
- [x] `npm run build` ✓ (dist 517.77 KB, −1.95 KB)

Closes the governance gap behind #223/#6; supersedes the budget logic in unmerged #7/#9/#194/#217 (their branches can be closed after this merges).


ranxianglei:
Follow-up (review nit 1 & 4): commit 8f390f2 removes the duplicated magic strings — `lastUserMessageId()` now matches against `SYNTHETIC_USER_PREFIXES = [THROTTLE_KICK_SENTINEL, DELEGATE_NOTIFY_PREFIX]`, and the delegate header is built from the newly exported `DELEGATE_NOTIFY_PREFIX` (src/delegate-tool.ts). Single source of truth on both the injector side and the matcher side; the constant documents the contract that any new synthetic `sendUserMessage` site must register its prefix there. Behavior unchanged (typecheck/build/tests/CI green).

ranxianglei:
关闭原因：这个修复是错误的——它偏离了本项目的既定设计，不应合入。

此前 review 给出的结论（LGTM、“修复正确且必要”）是错误的，请勿作为参考。作为 review 方，我未能理解到设计层面的意图，只机械地验证了实现本身的行为，是我太愚蠢，没能理解到这一层。

PR #254, opened 2026-08-30
fix(compress): repair missing closing brace on last content entry

### LINKED ISSUE #253: compress tool: model emits array missing closing '}' before final ']' — misleading "must be an ARRAY" error, lenient parse doesn't repair one missing brace
## Summary

With Qwen-family models in non-strict tool-call mode, the `compress` tool can fail repeatedly with a misleading error (`content must be an ARRAY`) even though the model **did** emit an array. The actual defect: the model omits the closing `}` of the last entry object before the final `]`, so `JSON.parse` rejects the whole string.

## What happened

3 consecutive `compress` failures in one session, each returning:

```
no-valid-ranges: got a JSON-encoded string (1 salvageable entries) but content must be an ARRAY of entry objects with keys: startId, endId, summary
```

The model retried 3 times (restructuring: adding `topic`, strict 3-key entries, ...) and only succeeded on the 4th attempt by accident. The error text never pointed at the real problem.

## Root cause

The `content` argument tail looked like this in all 3 failures:

```
..."summary":"... saved at /home/dog/tmp/comments_early.json."]
```

i.e. a well-formed array whose **last entry object is missing its closing `}`** — `"` + `]` instead of `"` + `}]`.

`JSON.parse` reports `Expecting ',' delimiter` at the last character. The error surfaced by the tool (`content must be an ARRAY`) is inaccurate — the input *is* an array shape; it just has one missing brace. This misleads capable models into restructuring the payload instead of fixing the brace.

Evidence: all 3 failed payloads shared the identical tail defect; replaying each through `JSON.parse` reproduces the exact error; a deterministic tail-repair recovers **3/3** payloads with valid `startId`/`endId`/`summary`.

This class of output is common with Qwen-family models when a single string field grows very large (multi-KB summaries), especially in OpenAI-compatible tool-call mode where arguments may arrive as a JSON-encoded string.

### 1. Deterministic tail-repair in `parseContentArray` (lenient parse)

If parsing fails and the trimmed input ends with `"]`, retry with `"}]`:

```js
function tailRepair(s) {
  const t = s.trimEnd();
  if (!t.endsWith("]")) return void 0;
  const body = t.slice(0, -1).trimEnd();
  if (!body.endsWith('"')) return void 0;
  try {
    const v = JSON.parse(body + "}]");
    if (Array.isArray(v)) return v;
  } catch {}
  return void 0;
}
```

Call it before `salvageContentEntries` when `JSON.parse` of the array fails. Repairs the missing-`}` case exactly; falls through untouched otherwise.

### 2. Accurate error message

When the input is already array-shaped but fails to parse, report the parser's own diagnostics instead of "must be an ARRAY", e.g.:

```
content array failed to parse: Expecting ',' delimiter near char 6051
```

This gives the model a retryable signal instead of sending it restructuring.

## Environment

- pi `0.1.4x`, tool-call via OpenAI-compatible endpoint (arguments delivered as JSON-encoded string)
- Model: Qwen-family 30B-A3B class, long single-string summaries (~6KB)

ranxianglei:
## Summary

Fixes #253.

With Qwen-family models in non-strict tool-call mode, the `compress` tool's `content` argument arrives as a JSON-encoded array string whose **last entry object is missing its closing `}`** (the tail is `\"` + `]` instead of `\"` + `}]`). `JSON.parse` rejects the whole string, and the tool surfaced a misleading error — `content must be an ARRAY` — even though the input *is* an array. This sent capable models into restructuring the payload instead of fixing the brace.

## Changes (`src/compress-tool.ts` only)

1. **Deterministic tail-repair** — `tailRepair()` / `repairContentTail()`: when `args.content` is a string that fails to parse and its trimmed form ends with `\"]` (preceded by `\"`), re-append `}` before the final `]` and retry. A no-op for well-formed input (a valid array never still parses after appending `}`), so there are no false positives. Wired into `normalizeRanges` before `parseCompressArgs`.
2. **Accurate error message** — `describeDiagnostics()` now detects array/object-shaped-but-unparseable content and reports the parser's own diagnostic (`the JSON failed to parse: Expecting ',' delimiter near char N`) instead of `must be an ARRAY`, giving the model a retryable signal. The `must be an ARRAY` fallback is preserved for genuinely non-array-shaped garbage.

## Why here and not in acp-kernel

The issue suggested the repair in acp-kernel's `parseContentArray`. I placed it in the adapter instead so this is a **self-contained, single-repo fix** — an acp-kernel change would require a cross-repo release (human PR merge + npm publish) before it could ship. The tail-repair could still be upstreamed to acp-kernel later so other hosts (opencode-acp) benefit; happy to open that as a follow-up if you want.

## Verification

- New `tests/compress-tail-repair.test.ts` (11 tests): single & multi-entry recovery, no-false-positive cases, trailing-whitespace tolerance, accurate error message, and an end-to-end tool test reproducing the #253 payload (single ~6 KB entry with the dropped `}`) that previously rejected with `no-valid-ranges … must be an ARRAY` and now resolves successfully.
- `npm run typecheck` clean · `npm test` **439 pass, 0 fail** (no regressions) · `npm run build` clean (dist 521.71 KB, repair present in bundle).

## Notes

- Content commit — no `version` bump (release-branch territory).
- Not merging (human-only).

### DIFF
--- a/src/compress-tool.ts
+++ b/src/compress-tool.ts
@@ -74,13 +74,44 @@ type RangeEntry = Static<typeof RangeSpec>;
 // the failure cap (a returned string would land as isError:false and count
 // as neutral). An empty array passes through (the call site returns "No
 // ranges provided.").
-function normalizeRanges(args: CompressArgs): RangeEntry[] | string {
-  const { ranges, diagnostics } = parseCompressArgs(args);
+export function normalizeRanges(args: CompressArgs): RangeEntry[] | string {
+  const effective = repairContentTail(args);
+  const { ranges, diagnostics } = parseCompressArgs(effective);
   if (ranges.length === 0) {
-    if (Array.isArray(args.content) && args.content.length === 0) return [];
-    return describeDiagnostics(diagnostics, args.content);
+    if (Array.isArray(effective.content) && effective.content.length === 0) return [];
+    return describeDiagnostics(diagnostics, effective.content);
   }
   return ranges.map((r) => ({ startId: r.startRef, endId: r.endRef, summary: r.summary, topic: r.topic }));
+}
+
+// Qwen-family models in non-strict tool-call mode sometimes emit the `content`
+// array as a JSON-encoded string whose LAST entry object is missing its closing
+// `}` (tail `"]` instead of `"}]`). The kernel's lenient parser then drops that
+// last range — or every range, when it is the only one — and reports a
+// misleading "truncated"/"no-valid-ranges" diagnostic. Repair the brace before
+// delegating so the whole array parses. Args are returned unchanged when the
+// repair does not apply.
+function repairContentTail(args: CompressArgs): CompressArgs {
+  if (typeof args.content !== "string") return args;
+  const repaired = tailRepair(args.content);
+  return repaired === undefined ? args : { ...args, content: repaired };
+}
+
+// Deterministic tail-repair: if the trimmed string ends with `"]` and the char
+// before it is a closing `"`, retry the parse with `"}]` appended. A valid JSON
+// array never still parses after appending `}`, so this has no false positives.
+export function tailRepair(s: string): string | undefined {
+  const t = s.trimEnd();
+  if (!t.endsWith("]")) return undefined;
+  const body = t.slice(0, -1).trimEnd();
+  if (!body.endsWith('"')) return undefined;
+  const candidate = body + "}]";
+  try {
+    if (Array.isArray(JSON.parse(candidate))) return candidate;
+  } catch {
+    // not the missing-brace case
+  }
+  return undefined;
 }
 
 function describeDiagnostics(diagnostics: CompressParseDiagnostics, content: CompressArgs["content"]): string {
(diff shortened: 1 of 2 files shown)


ranxianglei:
> 🤖 ework agent · vllm-qwen/qwen3.8-27b

[bot] **Reviewed & verified — ready to merge** (merge left to you, human-only).

I didn't just trust the PR description — I re-ran everything on the branch (`2026-08-30_compress-tail-repair-missing-brace`, commit `812ddc0`) and added a before/after reproduction to prove the new test is meaningful.

### Verification (all green, matches the claims)
- `npm run typecheck` — clean
- `npm test` — **439 pass, 0 fail** (incl. the 11 new `tests/compress-tail-repair.test.ts`)
- `npm run build` — clean, `dist/index.js` 521.71 KB, repair code present in the bundle

### Before/after (the part I added)
Replayed the exact #253 payload (single entry, ~6 KB summary, dropped closing `}`) through the real `compress` tool `execute` path on **both** trees:
- **master**: rejects with `Invalid compress content (no-valid-ranges): … content must be an ARRAY of {startId, endId, summary} objects.` — the misleading error, even though the input *is* an array.
- **PR branch**: the same payload is repaired and compresses successfully (returns the `▣ ACP |` panel, no parse error).

So the e2e test genuinely reproduces the bug and the fix resolves it — not passing vacuously.

### Code review (traced the logic, no issues)
- `tailRepair` is sound and **false-positive-free**: a valid object array ends in `}]` (body ends in `}`, not `"`) so it never reaches the parse retry; string arrays / empty arrays / non-array shapes / mid-array defects all fall through untouched. I traced each case.
- `repairContentTail` is pure — only touches `string` content, returns args unchanged when the repair doesn't apply, so well-formed input is byte-for-byte unaffected.
- `describeDiagnostics` ordering is correct: `invalidItems` first, then the new `jsonParseError` (parser's own diagnostic), then the `must be an ARRAY` fallback. A genuinely non-array shape (e.g. a single JSON object, or `[1,2,3]`) still gets the right message.
- `normalizeRanges` is now `export`ed for testability — acceptable.
- Test conventions match the repo (harness `any` types, `/tmp/pai-acp-*` state files are established patterns).

### Notes
- Content commit, no `version` bump — correct per spec.
- Upstreaming the tail-repair to acp-kernel's `parseContentArray` later (so opencode-acp benefits) is a sensible optional follow-up — happy to open that issue if you want it tracked.

LGTM. Merging is yours.

PR #237, opened 2026-08-26
refuse ACP service on OMP hosts (oh-my-pi)

ranxianglei:
## Summary

Refuses ACP service on **OMP** (`can1357/oh-my-pi`) hosts. Fixes #234.

## Why

OMP's in-process session API diverges from Pi's. The compression refs billion-context-pi injects can drift out of sync with the session's real refs, so `compress` calls fail with `does not exist in this session` and one-shot runs never succeed. OMP is no longer actively supported on this plugin (maintainer note on #234) — the in-process integration is too complex to keep reliable.

## What changed

- **Host detection** — new `src/omp.ts`: `isOmpHost(sm) = !isPiHost(sm)`, reusing the existing feature-detect (Pi exposes `buildContextEntries`; OMP only `getBranch`).
- **Refuse at `session_start`** — on OMP the extension sets a `refused` flag and, once per process, warns (UI `notify` in TUI/RPC; `console.error` to stderr in headless one-shot; plus a log line) pointing users at the supported alternative.
- **Stand down** — when refused: the 4 ACP tools return the guidance message, the ACP system-prompt injection is skipped, the `context` transform is a no-op (messages pass through untouched), and host auto-compaction is left alone (no `cancel`).
- **Pi hosts are unaffected.**
- **Docs** — README gains a "Host support" section recommending [billion-context](https://github.com/ranxianglei/billion-context) for OMP.

## Alternative for OMP users

```bash
npm install -g billion-context
bili omp -- -p "your task"   # OMP through the proxy — refs never diverge
```

## Tests

- New `tests/omp-refuse.test.ts` (5 cases): single warning, tool refusal, headless stderr path, and Pi-host no-regression.
- `npm test` → **433 pass, 0 fail**. The existing OMP integration tests still pass (they exercise the context handler directly without firing `session_start`).
- `npm run typecheck` clean; `npm run build` succeeds.

No version bump here — that belongs to the release commit.

PR #256, opened 2026-08-30
Show /acp and /acp-status as persistent custom messages (#255)

ranxianglei:
## Problem (#255)

In web hosts such as agegr/pi-web, `/acp` and `/acp-status` output goes through `ctx.ui.notify()`, which renders as a transient toast — the status panel disappears and is not part of the session transcript.

## Root cause

Not a pi-web bug: pi-web renders `notify()` as a toast by design (that's the documented semantics of the notify channel) and renders `custom_message` session entries persistently (`CustomMessageView` in the transcript). The bug is on our side — billion-context-pi routes persistent status content through the transient notification channel.

## Fix

- `/acp` and `/acp-status` now emit `pi.sendMessage({ customType: "acp-status", content: text, display: true })` — the documented mechanism for extension-injected messages. This persists a `custom_message` entry in the session JSONL (idle and streaming paths both persist via the session manager's `message_end` handler) and renders persistently in the TUI (`CustomMessageComponent`) and in web hosts (pi-web's `CustomMessageView`).
- `ctx.ui.notify(text)` remains the fallback for hosts without `sendMessage` (older pi).
- `entriesToCoreMessages` skips `acp-status` custom messages: the panel is UI-only and must never be projected into the sent view (otherwise every `/acp` run would add 1-3K tokens to the model's context). The model already has the `acp_status` tool for programmatic status.

## Tests

3 new tests (431 total, all passing):
- both commands emit the panel as a persistent custom message with `display: true`, no notify
- fallback to notify when `pi.sendMessage` is absent
- `entriesToCoreMessages` excludes `acp-status` panels while other extensions' custom messages still project

`npm run typecheck && npm test && npm run build` all green.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -1,5 +1,6 @@
-import type { ExtensionCommandContext, RegisteredCommand, SessionEntry } from "@earendil-works/pi-coding-agent";
+import type { ExtensionAPI, ExtensionCommandContext, RegisteredCommand, SessionEntry } from "@earendil-works/pi-coding-agent";
 import type { AcpRuntime } from "./runtime.js";
+import { ACP_STATUS_CUSTOM_TYPE } from "./messages.js";
 import { defaultCountTokens, parseBlockIdArg, collectBlockContent } from "acp-kernel";
 import { getSystemPromptText } from "./compat.js";
 import { collectCoveredMessageIds, estimateTokens, calibrateTokens, collectImageTokens, modelSupportsImages } from "./tokens.js";
@@ -25,20 +26,31 @@ function cacheUsageSamples(entries: SessionEntry[]): Array<{ input: number; cach
   return out;
 }
 
-export function makeCommands(runtime: AcpRuntime): Array<{ name: string; options: CommandOptions }> {
+export function makeCommands(runtime: AcpRuntime, pi?: ExtensionAPI): Array<{ name: string; options: CommandOptions }> {
+  // Persistent transcript output (rendered by TUI and web hosts like pi-web);
+  // notify() is a transient toast and only the fallback for hosts without
+  // sendMessage (issue #255).
+  const statusHandler = async (_args: string, ctx: ExtensionCommandContext) => {
+    const text = await statusReport(runtime, ctx);
+    if (typeof pi?.sendMessage === "function") {
+      pi.sendMessage({ customType: ACP_STATUS_CUSTOM_TYPE, content: text, display: true });
+      return;
+    }
+    ctx.ui.notify(text);
+  };
   return [
     {
       name: "acp",
       options: {
         description: "Show ACP context usage, token breakdown, and compression status.",
-        handler: async (_args, ctx) => ctx.ui.notify(await statusReport(runtime, ctx)),
+        handler: statusHandler,
       },
     },
     {
       name: "acp-status",
       options: {
         description: "Detailed ACP status (block tiers, token breakdown, delegate usage).",
-        handler: async (_args, ctx) => ctx.ui.notify(await statusReport(runtime, ctx)),
+        handler: statusHandler,
       },
     },
     {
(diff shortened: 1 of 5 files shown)


ranxianglei:
> 🤖 ework agent · vllm-qwen/qwen3.8-27b

[bot] Reviewing PR #256 now — diff is small (5 files, +78/−8). I've fetched the branch and am running `npm run typecheck && npm test && npm run build` to verify the claims, plus checking the `pi.sendMessage` contract against the installed pi types. Will report findings shortly.

ranxianglei:
> 🤖 ework agent · vllm-qwen/qwen3.8-27b

[bot] ## Review: PR #256 — /acp & /acp-status as persistent custom messages

**Verdict: LGTM** — all claims verified, one minor gap noted below (non-blocking for the pi-web target).

### Verified against pi 0.83.0 (installed `@earendil-works/pi-coding-agent`)

1. **CI green, matches claims.** `npm run typecheck` clean, `npm test` 431/431 passing, `npm run build` OK (dist/index.js 520.92 KB).
2. **`pi.sendMessage` contract matches exactly.** `ExtensionAPI.sendMessage<T>(message: Pick<CustomMessage<T>, "customType" | "content" | "display" | "details">, options?: { triggerTurn?: boolean; deliverAs?: ... }): void` (`dist/core/extensions/types.d.ts:911`). No options passed ⇒ no turn triggered. ✓
3. **Persistence claim holds.** `sendCustomMessage` (`dist/core/agent-session.js:1065`): idle path pushes `{ role: "custom", ... }` to `agent.state.messages` **and** calls `sessionManager.appendCustomMessageEntry(...)`; streaming routes via the steer/followUp queue and persists on delivery. ✓
4. **TUI renders persistently.** `addMessageToChat` (`dist/modes/interactive/interactive-mode.js:2628`): `role === "custom"` with `display: true` → `CustomMessageComponent` added to the persistent chat container (purple box, `[acp-status]` label, markdown content). Not a toast. ✓
5. **Sent-view exclusion is necessary and correct on pi hosts.** pi's own `sessionEntryToContextMessages` converts `custom_message` entries into user messages in `buildSessionContext()` by default — so without the filter, every `/acp` run would pollute the model context. On pi hosts, `stateFor` reads entries via `buildContextEntries()` (which includes `custom_message` entries), and the PR's filter in `entriesToCoreMessages` (`entry.type === "custom_message" && entry.customType !== ACP_STATUS_CUSTOM_TYPE`) skips the panel. Other extensions' custom messages still project. ✓

### One gap: omp (non-pi) host leak — minor, optional follow-up

On non-pi hosts, `stateFor` takes the live-merge path (`src/runtime.ts:378`, `!isPiHost(sm) && liveMessages?.length`). There:

- `mergeLiveEntries` (`src/runtime.ts:106-134`) builds output **only from live messages**; persisted `custom_message` entries are excluded from identity matching (`persisted = entries.filter(e => e.type === "message")`), so the live `role: "custom"` panel message becomes a plain `{ type: "message", message }` entry.
- `projectMessage` (`src/messages.ts:45-93`) projects any unknown role (including `"custom"`) with non-empty text as a **user message** — so on omp the panel would leak into the sent view, and the PR's filter never fires (it only matches `entry.type === "custom_message"`).

Severity: minor. The PR's actual target (pi-web/TUI) is fully fixed; omp would need to implement `sendMessage` for this path to trigger, and #237 (refuse ACP service on OMP hosts) is already open. Suggested hardening if you want it: also drop `role: "custom"` messages with `customType === ACP_STATUS_CUSTOM_TYPE` in `projectMessage` — a one-line guard that covers both paths. Happy to take that as a follow-up PR if you'd like.

Otherwise the design is right: `sendMessage` is the documented mechanism, `notify` stays as the fallback for older hosts, and the UI-only panel is correctly kept out of the model context on pi hosts.

PR #270, opened 2026-09-01
feat(compress): server-side auto-compress enforcement when the model ignores nudges (closes #269)

### LINKED ISSUE #269: Title: Context never auto-compresses when the model ignores the nudge — climbs 75%→95%+ on long agentic tasks (no enforcement path)
## Summary                                                                                                                                                                                                            
   billion-context-pi only ever TRIGGERS compression by injecting an advisory                                                                                                                                            
   nudge prompt; the actual compression requires the model to call the `compress`                                                                                                                                        
   tool itself. There is no enforcement / auto-compress path. When the model is                                                                                                                                          
   mid-way through a long agentic task and ignores the nudge, context grows                                                                                                                                              
   unbounded past the 75% OVER-LIMIT threshold all the way to 95%+ (EMERGENCY),                                                                                                                                          
   and EMERGENCY is also only a stronger label + output-headroom shrink                                                                                                                                                  
   (262144→245760) + tool-output truncation — it still never force-compresses.                                                                                                                                           
                                                                                                                                                                                                                         
   ## What 0.1.55 fixed (works)                                                                                                                                                                                          
   The METER / detection: tokenCount = max(estimate, providerReal), so the                                                                                                                                               
   OVER-LIMIT nudge now fires at true 75% usage. Confirmed in acp.log —                                                                                                                                                  
   session 01a01e31 on 0.1.55 logs `nudge=active nudgeReason=OVER-LIMIT T1,                                                                                                                                              
   usage 77%` on EVERY turn from 77% up to 87% for 20+ consecutive turns.                                                                                                                                                
                                                                                                                                                                                                                         
   ## What is still broken (the ask)                                                                                                                                                                                     
   1. No enforcement. Over 20 consecutive OVER-LIMIT turns the model did not                                                                                                                                             
      call compress (only one earlier compliant compress: 184K→112K). There is no                                                                                                                                        
      fallback that compresses server-side when the model keeps ignoring.                                                                                                                                                
   2. Tiny compressible headroom during active work. At 87% usage the log shows                                                                                                                                          
      `max effective pending 16296` (est) — most of the context is either in the                                                                                                                                         
      protected recent/current-step zone or already in 71 blocks, so even a                                                                                                                                              
      compliant compress only reclaims ~16K (87%→~80%) and the running task                                                                                                                                              
      regrows it within ~30 min. The protected "recent/last-user zone" may be too                                                                                                                                        
      large for long single-step agentic runs.                                                                                                                                                                           
                                                                                                                                                                                                                         
   ## Suggested fix                                                                                                                                                                                                      
   Add an enforcement path: after N consecutive ignored OVER-LIMIT/EMERGENCY                                                                                                                                             
   nudges (or hard at 0.95), the extension should auto-compress the largest                                                                                                                                              
   compressible ranges itself (server-side), independent of the model. Also                                                                                                                                              
   consider bounding the protected recent-zone by TIME/TOKENS (not "since last                                                                                                                                           
   user message") so long agentic steps don't pin ~all context as non-compressible.                                                                                                                                      
                                                                                                                                                                                                                         
   ## Repro / evidence                                                                                                                                                                                                   
   - pi 0.83.6, billion-context-pi 0.1.55, local sglang Qwen3.8-27B                                                                                                                                                      
     (window 262144→245760), CJK-heavy.                                                                                                                                                                                  
   - acp.log [turn] for sid 01a01e31, 07:57–08:06Z:                                                                                                                                                                      
     `nudge=active nudgeReason=OVER-LIMIT T1: max effective pending 16296,                                                                                                                                               
      usage 87% blocks=71 activeBlocks=17`  (repeated; no [compress] events).                                                                                                                                            
   - One compliant compress at 07:25Z: `applied ranges=3 tokensCompressed=29775                                                                                                                                          
     beforeTokens=37660 afterTokens=9148` then regrew 112K→212K in ~30 min.

ranxianglei:
## Problem

Nudges are **advisory** — they only work if the model actually calls `compress`. On long agentic tasks a model can ignore them for 20+ consecutive over-limit turns (77%→87% in the issue log) with **no server-side fallback**, climbing toward overflow. There was no enforcement path.

## Fix

Add a server-side enforcement path (`src/auto-compress.ts`) that auto-compresses the largest compressible ranges **itself**, independent of the model, in two ways:

- **Streak** — after `afterIgnores` (default 3) consecutive over-limit turns where the model did not compress.
- **Hard threshold** — once usage reaches `hardThreshold` (default 95%), immediately.

### Design

- Runs in the `context` transform right after `processTurn`, before nudge injection — so the model sees the post-compression context and the nudge reflects the new (lower) usage.
- `pickAutoRanges` greedily selects viable ranges (largest first) until projected usage drops below `targetPct` (default 80%), capped by `maxRanges` (default 5).
- Auto-compressed ranges get an **honest, labeled mechanical summary** (the block is fully restorable via `decompress`) — no fabricated model-quality summary.
- A successful model-driven `compress` resets the streak; enforcement never fires below `targetPct`; a once-per-turnKey gate (`lastAutoCompressedTurnKey`) prevents same-turn spam; `blocksCreated === 0` is handled gracefully.
- `compress.autoCompress` defaults to **`true`** (it's the bug fix) and is disable-able via `false` or an object with `afterIgnores` / `hardThreshold` / `targetPct` / `maxRanges`.

No acp-kernel change needed — all required APIs (`applyCompression`, `processTurn`, `nudge.compressibleRanges`, `viableRanges`) already exist in 0.0.48.

## Tests

12 new tests in `tests/auto-compress.test.ts`:
- **Unit:** `resolveAutoCompress` (defaults / false / object + percent strings), `AutoCompressEpisode.reset`, `pickAutoRanges` (greedy + maxRanges cap), `buildHeuristicSummary` (shape + snippet truncation + 12K cap), `messagesInRange`.
- **E2E:** streak trigger (block only after 3 ignored nudges), hard threshold (block on first fire at 96%), disabled (no blocks with `autoCompress: false`).

Full suite: **474 pass / 0 fail / 3 skipped** (pre-existing). `typecheck` clean, `build` succeeds.

## Docs

- `CONFIGURATION.md` + `CONFIGURATION.zh-CN.md`: new `compress.autoCompress` reference (summary table, flow tier 4, detailed subsection with field table + example).
- `README.md`: one-paragraph surface-level note.
- `CHANGELOG.md`: Unreleased entry.

Closes #269.


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewing PR #270 (server-side auto-compress enforcement, closes #269). I'm checking out the branch, reading `src/auto-compress.ts` + the wiring in `index.ts`/`runtime.ts`/`config.ts`, and running the full suite (typecheck / test / build) to verify the claims. I'll report findings here — will flag any issues before you merge.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review — PR #270** (server-side auto-compress enforcement, closes #269)

**Verdict: ready to merge.** I checked out the branch and verified every claim. Found **3 minor, non-blocking notes** — one contradicts a stated design goal and is worth a quick follow-up.

**Verification (ran on `2026-09-01_auto-compress-enforcement`)**
- `npm run typecheck` — clean
- `npm test` — **474 pass / 0 fail / 3 skipped** (matches)
- `npm run build` — succeeds (dist/index.js 549 KB, acp-kernel inlined)
- Kernel API usage — correct against acp-kernel 0.0.48: `applyCompression`/`processTurn`/`viableRanges`/`refToIndex`, `state.messageRefs.byRaw`, and `nudge.{shouldInject, breakdown.overLimit, breakdown.emergencyOverride, compressibleRanges}` all match the shipped types.
- Wiring — moving `turnKey`/`compressOutcomes`/`noteCompressOutcomes` **before** `processTurn` is safe: all three depend only on `entries`, not on `turn` (noteCompressOutcomes is idempotent per toolCallId). Episode lifecycle (`autoCompressFor`/`autoCompressDrop`, reset on `session_start`, drop on `session_shutdown`) mirrors the existing overflow/throttle episode pattern.

**Design — sound.** Runs after `processTurn`, before nudge injection; greedy largest-first range selection down to `targetPct`; honest, labeled mechanical summary (block fully restorable via `decompress`); streak reset on a successful model `compress`; `usage <= targetPct` guard; once-per-turnKey gate; `blocksCreated === 0` handled. Config surface (`false` / object with per-field merge + percent-string parsing) is clean and well-tested.

**Minor notes (non-blocking):**

**1. Post-compress nudge is stale — contradicts a stated goal.** `maybeAutoCompress` re-runs `processTurn({ …, tokenCount })` with the **pre-compression** `tokenCount` (src/auto-compress.ts:242). The kernel computes nudge `usage = tokenCount / modelContextLimit`, so the returned nudge's `contextUsage`/`overLimit`/`emergencyOverride`/`shouldInject` — and the `logInfo("turn")` `pct` — reflect the **pre-compression (higher)** usage, not the lower post-compression usage. The PR says "the nudge reflects the new (lower) usage"; that half isn't met (the *messages* are post-compression, which is the part that matters). Impact is contained: it self-corrects on the next context event (which recomputes `tokenCount` from the compressed session log), and the once-per-turnKey gate prevents a re-fire — worst case is one misleading over-limit/emergency nudge in the event where enforcement fired. Suggested fix: recompute `tokenCount` from the post-compression messages before the re-run `processTurn` (or suppress the nudge for that single event).

**2. Streak is counted per context-event, not per user-turn — docs say "turns".** The code comment reads "Consecutive over-limit context events" and the e2e test fires 3 context events within a single `turnKey`. But the PR description + `CONFIGURATION.md` both say "consecutive over-limit **turns**". Since Pi fires the `context` event multiple times per assistant reply (tool loop), `afterIgnores=3` can be reached **within one user turn**, making the trigger more aggressive than "3 turns". It's bounded (only fires above `targetPct`=80%, at most once per turnKey), so practical impact is low — but the docs should say "context events" (or the counter should be keyed per turnKey) to match the code.

**3. Heuristic summary can exceed a small CJK-heavy range (net token loss).** The summary is capped at 12K *chars*; for CJK (~1 token/char) a viable range just above `minCompressRange` (5000 chars) with many short messages can yield a summary *larger* than the range, so that range's compression **increases** context. Mitigated by largest-first ordering (big ranges net-gain), the `targetPct` guard, `maxRanges`, and the `blocksCreated===0` path — but consider capping the summary relative to the range's token size (or skipping ranges where the projected summary ≥ the range).

**On issue #269 point 2** (bounding the protected recent-zone by time/tokens so long single-step agentic runs don't pin most of the context as non-compressible): correctly deferred — it's a kernel-side knob (`preserveRecentTokens`/`preserveRecentMessages`), so it belongs in `acp-kernel` first. Good scoping.

**Recommendation:** merge is yours. I'd address **#1** (contradicts the stated design) and **#2** (doc accuracy) — both small — either before merge or as a quick follow-up; #3 is a nice-to-have. Happy to push fixes for #1/#2 if you want. (Per repo rules I won't merge the PR myself.)

PR #293, opened 2026-09-04
fix(tokens): measure tokenCount on the actual sent view (#289)

ranxianglei:
## Problem

#289 reports token counts desyncing from post-compression state, causing consecutive low-yield / zero-yield compressions. All four sub-claims were verified against the code and the incident logs:

1. **`tokens` / `pct` / `usage` disagree** — two mechanisms:
   - The `[turn]` log mixed scales: `tokens` = ACP local estimate, `pct` = Pi host `getContextUsage().percent` (host-window scale). The incident math proves it: `pct=19.970955882352943 × 272,000 = 54,321` exactly equals the host-reported token count, while `tokens=190,513`.
   - **Root cause**: the base estimate `estimateTokens(coreMessages, coveredIds)` runs over the RAW core view minus block coverage — not the sent view `processTurn` actually produces (prune + summary injection + orphan-strip + absorb + filter). In long multi-block sessions these diverge without bound: content that is uncovered in the raw view but stripped from the sent view every turn (e.g. tool results whose paired call got covered) is counted forever. Incident: local estimate 190,513 vs real provider input 36,387 (4,899 input + 31,488 cacheRead) — a ~174K constant offset that tracks compression exactly (next turn: 183,875; Δ6,638 ≈ tokensCompressed 6,717).
2. **`sentTokens` not recomputed after compression** — confirmed: the second `processTurn` after `applyCompression` reused the pre-compression `sentTokens`, so the emergency-truncation band and the panel "reclaimed" number were computed on a stale scale.
3. **Tiny-pending T2/T3 distill still firing EMERGENCY** — the kernel pressure band injects the max-pending tier whenever `pending > 0` (no minimum-benefit gate). With claim-1's inflated count pinning usage ≥ 95%, this loops: each micro-compression succeeds → nudge baselines reset → next LLM call re-injects. Downstream symptom of #1; the gate itself belongs in acp-kernel → companion issue filed there.
4. **Zero-yield compressions not stopping** — the per-turn cap (`MAX_COMPRESS_ATTEMPTS=3`) exists and counts failures/no-ops, but resets every user turn; with the count pinned high the loop persisted across turns. Removing the false emergency (claim 1) is what actually stops it.

## Fix (adapter-side, no acp-kernel bump)

- **`src/tokens.ts`**: new `sentViewTokenCount()` — runs `processTurn` on a throwaway `structuredClone(state)` and re-measures tokens on the actual post-pipeline view (the clone is mandatory because `processTurn` mutates state); `adjustedTokenCount()` adopts the recount when coverage exists and the delta exceeds `max(1000, 10%)`.
- **`src/index.ts`** (Fix A): `wireContextTransform` recounts before the final pass when active blocks exist; all floors stay raise-only. (Fix C) the `[turn]` log now computes `pct` from the same `tokenCount/limit` scale and logs host usage separately as `hostTokens`/`hostPct`, appended after `activeBlocks` to keep the existing field layout stable for log consumers.
- **`src/compress-tool.ts`** (Fixes A′/B): the first pass uses the recounted count; the post-compression `afterTurn` now measures from `applied.state` instead of the stale pre-compression `sentTokens`.

## Tests

- New `tests/view-recount.test.ts`: an e2e discriminator (one assistant turn with 21 parallel tool calls; c1's result sits 21 positions past c1's split core — beyond `adjustBoundariesForToolPairs`' `maxScan=20` — so compressing c1 alone orphans its result. Raw view then reads ~97% ≥ emergency band while the honest sent view reads ~72%; asserts no EMERGENCY nudge fires) plus a helper sanity test (zero-block state does not drift). Fails on pre-fix code, passes with the fix.
- Full suite: 467 tests, 464 pass, 0 fail, 3 skipped; typecheck + build green.

## Expectation mapping from #289

| # | Expected behavior | Resolution |
|---|---|---|
| 1 | compute tokens from the final sent list | ✅ Fixes A/A′ |
| 2 | recompute postCompressionTokens before processTurn | ✅ Fix B |
| 3 | minimum-benefit gate for T2/T3 nudges | → acp-kernel companion issue (cross-repo release ordering) |
| 4 | stop after a zero-reclaim compress within a turn | already present (per-turn cap, verified); the false emergency is what let it loop across turns |
| 5 | cross-turn circuit breaker | declined: fixing the count removes the spurious permanent emergency; an extra breaker risks suppressing legitimate compression |

No `version` bump (release-branch territory). acp-kernel stays pinned at exact 0.0.48.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -3,7 +3,7 @@ import type { AcpRuntime } from "./runtime.js";
 import { ACP_STATUS_CUSTOM_TYPE } from "./messages.js";
 import { defaultCountTokens, parseBlockIdArg, collectBlockContent } from "acp-kernel";
 import { getSystemPromptText } from "./compat.js";
-import { collectCoveredMessageIds, estimateTokens, collectImageTokens, modelSupportsImages } from "./tokens.js";
+import { collectCoveredMessageIds, estimateTokens, collectImageTokens, modelSupportsImages, adjustedTokenCount } from "./tokens.js";
 import { usageAnchorPredatesCompression } from "./floor-stale.js";
 import { buildStatusPanel } from "acp-kernel/panel";
 import { getDelegateUsage } from "./delegate-tool.js";
@@ -143,9 +143,13 @@ async function statusReport(runtime: AcpRuntime, ctx: ExtensionCommandContext): 
   const sessionTokens = !anchorStale && realUsage?.tokens && realUsage.tokens > 0 ? realUsage.tokens : defaultCountTokens(coreMessages.map((m) => m.text ?? "").join("\n")) + imageTokensTotal;
   const coveredIds = collectCoveredMessageIds(state);
   const sentTokens = estimateTokens(coreMessages, coveredIds, imageTokens) + systemPromptTokens;
+  // View-based recount (issue #289): with active blocks the raw-view estimate
+  // can sit far above the sent view and mis-scale the panel's nudge — same
+  // arbitration as src/index.ts and acp_status.
+  const viewSentTokens = adjustedTokenCount(runtime.core, coreMessages, state, config, sentTokens, imageTokens, systemPromptTokens);
   // issue #257: floor the meter at the host's real context usage so the
   // panel's nudge matches the real decision (same as src/index.ts).
-  const turn = runtime.core.processTurn({ messages: coreMessages, state, config, tokenCount: anchorStale ? sentTokens : Math.max(sentTokens, realUsage?.tokens ?? 0) });
+  const turn = runtime.core.processTurn({ messages: coreMessages, state, config, tokenCount: anchorStale ? viewSentTokens : Math.max(viewSentTokens, realUsage?.tokens ?? 0) });
 
   // Shared kit surface renders the panel (dual accounting, viability
   // filtering, bars, block list with topic fallback). Host-specific inputs:
--- a/src/compress-tool.ts
+++ b/src/compress-tool.ts
@@ -7,7 +7,7 @@ import type {
 import type { AcpRuntime } from "./runtime.js";
 import { MAX_COMPRESS_ATTEMPTS } from "./runtime.js";
 import { debug, logError, logInfo, logThrow, logWarn } from "./log.js";
-import { estimateTokens, collectCoveredMessageIds, collectImageTokens, modelSupportsImages, lastUserMessageId } from "./tokens.js";
+import { estimateTokens, collectCoveredMessageIds, collectImageTokens, modelSupportsImages, lastUserMessageId, adjustedTokenCount } from "./tokens.js";
 import { defaultCountTokens, parseCompressArgs, viableRanges, formatRanges, type CompressionBlock, type CompressionState, type CompressParseDiagnostics, type NudgeDecision } from "acp-kernel";
 import { getSystemPromptText } from "./compat.js";
 import { OMP_UNSUPPORTED_MESSAGE } from "./omp.js";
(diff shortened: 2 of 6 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review of PR #293 — LGTM ✅ (verified green across local + CI)**

I reviewed the diff (`src/tokens.ts`, `src/index.ts`, `src/compress-tool.ts`, `tests/view-recount.test.ts`; 4 files, +214 −28) and independently re-ran the gates rather than trusting the PR body.

**Verification (all pass):**
- typecheck: exit 0
- build: `dist/index.js` self-contained (acp-kernel inlined)
- full suite: **467 tests / 464 pass / 0 fail / 3 skipped** — matches the PR claim exactly
- new `tests/view-recount.test.ts`: both cases pass
- GitHub CI on head `b280b76`: all 9 checks green (test × ubuntu/windows × Node 22/24, e2e, docker, pr-validation, build-artifact); PR is `mergeable_state: clean` vs master

**Correctness — the fix targets the root cause, not the symptom:**
- **#289 claim 1 (root):** `estimateTokens(coreMessages, coveredIds)` counted uncovered messages that prune strips from the sent view every turn (orphaned tool pairs stranding past a block boundary), pinning usage in the emergency band → low/zero-yield loops. `sentViewTokenCount()` (`src/tokens.ts:58`) re-measures on the actual post-`processTurn` view via a throwaway `structuredClone(state)` — the clone is load-bearing because `processTurn` mutates state. Adopted only when divergence exceeds `max(1000, 10%)`.
- **Claim 2 (Fix B):** the compress post-pass now recounts from `applied.state` (`src/compress-tool.ts:373`) instead of the stale pre-compression `sentTokens`, so the emergency-truncate band (0.95) and the "reclaimed" number sit on the right scale.
- **Floors refactor** (`src/index.ts:249-272`): `applyFloors()` is raise-only, captures the host-usage floor + armed self-heal floor once, and applies identical floors to whichever base wins; `ov.armed` is consumed exactly once either way. The recount is gated on active-block coverage, and the probe never mutates the real state.
- **`[turn]` log (Fix C):** `pct` is now computed on the single `tokenCount/limit` scale; host-reported usage is logged separately as `hostTokens`/`hostPct`, appended after `activeBlocks` to keep the pre-existing field layout stable for log consumers.

**Code quality:** no `as any` / `@ts-ignore` introduced; comments are rationale-bearing (justified here); the test file's `any` usage matches existing test conventions.

**Non-blocking observations (no separate issues warranted):**
1. **Kernel version drift.** The PR body says "acp-kernel stays pinned at exact 0.0.48," but master has since bumped to **0.0.50** (v0.1.56, unrelated). The branch is based on v0.1.55 and doesn't touch `package.json`, so a plain merge cleanly lands on 0.0.50. I verified the fix compiles **and** passes the full suite against **0.0.50 as well**, so it's functionally safe — but consider rebasing onto current master so pre-merge CI exercises the version that actually ships.
2. **Probe inherits the emergency-band truncation.** `sentViewTokenCount` probes with `tokenCount: prelim`; in the exact #289 scenario `prelim` sits ≥95%, so the probe itself fires emergency tool-result truncation and measures a prune+truncate view rather than the pure prune-only sent view → a slight under-count. Mitigated because `applyFloors` floors at host-reported real usage when available, and the primary goal (suppressing the false EMERGENCY) holds either way. Optional tightening: probe with a sub-band tokenCount to isolate the prune view.
3. **Perf tradeoff.** Every context turn with ≥1 active block-with-coverage now runs an extra full `processTurn` (the probe); the compress path adds up to two more. That's a steady ~2× pipeline cost per turn in multi-block sessions — acceptable for correctness; an optional gate (only recount when the raw estimate nears a band) could trim it later.

**Bottom line:** correct, well-scoped, fully green. Per system rules I can't merge PRs (human-only) — please merge when ready: https://github.com/ranxianglei/billion-context-pi/pull/293

PR #304, opened 2026-09-06
feat(fleet): bordered full-width /acp-fleet panel with full session transcript

ranxianglei:
## Summary

Rewrites the `/acp-fleet` live inspector (option B) to fix the reported UX issues: background bleed-through, no visible frame, the transcript showing only a single pair of tools, and inconsistent scrolling.

## Changes

**Bordered, opaque full-width panel**
- The inspector now renders a box-drawing frame (`┌─┐│└┘`) at **100% width**, with every interior line space-padded so nothing from the underlying chat bleeds through. No more messy transparency — it reads as a clean framed modal.

**Full sub-session transcript**
- Previously the detail view tailed `<runId>.activity` (tool stream only — which is why it looked like "just one pair of tools"). It now parses `<runId>.session.jsonl` (the full pi session log, already exposed as `FleetRunView.sessionFile`) and renders the whole conversation:
  - `you ›` user messages
  - `⌥` assistant thinking blocks
  - assistant text
  - `🔧 tool {args}` calls and `✓/✗ tool: result` results (one-line preview)
- New pure helpers: `parseSessionJsonl()`, `renderTranscriptBlocks()`, `readHeadSync()` (read-from-start, complementing the existing `readTailSync`).

**Scrolling + live refresh**
- Both list and transcript scroll with **↑↓ / jk AND PgUp/PgDn**; the list auto-scrolls to keep the selection visible; the transcript follows the live tail until you scroll up.
- Kept the 500ms tick → `requestRender()` live refresh.

## Verification
- `tsc --noEmit` clean; full test suite green; build OK.
- Live tmux check (pi 0.83.6): list view shows a bordered full-width table; transcript view shows conversation + thinking + tool call/result for a real worker run; PageUp jumps to top, Down scrolls line-by-line.

## Notes
- Non-TUI hosts still get the plain-text snapshot (`buildSnapshotText`).
- Merge is human-only per AGENTS.md.


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewing #304 now. Status so far:

**Verification (on PR head `55041a1`, fresh clone + `npm ci`):**
- `npm run typecheck` — clean
- `npm test` — 544 pass / 0 fail / 3 skip (the 3 skips are pre-existing in unrelated modules: omp message-identity and live-context-ref tests; none introduced by this PR)
- `npm run build` — success (`dist/index.js` 677.93 KB)

So the "tsc clean / suite green / build OK" claim checks out.

**Code review (in progress):** I've verified every external API this PR leans on against the installed pi 0.83.0 packages — `theme.bg("selectedBg", …)` exists on the `Theme` class, `truncateToWidth(…, pad=true)` is ANSI-aware, and the session-jsonl field names (`thinking`, `toolCall.name/arguments`, `toolName`, `isError`, `model_change.provider/modelId`) match what pi's `session-manager` actually writes. Full write-up with any nits follows shortly.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review complete — LGTM, ready for human merge.** No blocking issues found. Details below.

## Duplicate screening
Only related items: #295 (original inspector, merged) and #303 (queued-icon fix, merged). This PR is a follow-up rewrite fixing post-#295 UX reports plus the new transcript feature — different scope, not a duplicate of either. No separate issue tracking these UX complaints exists, so nothing else to cross-link.

## Claim verification (all checked against installed pi 0.83.0 packages, not just the diff)
1. **Opaque bordered panel** ✅ — `frame()` space-pads every interior line to full width via ANSI-aware `truncateToWidth(…, pad=true)` (verified its impl in pi-tui `utils.js`: width computed on visible graphemes, not raw length), so colored selection bars can't bleed through. Overlay `width:"100%"` confirmed.
2. **"Just one pair of tools" root cause** ✅ correctly layered — old code tailed `<runId>.activity` (tool stream only); the fix swaps the data source to `<runId>.session.jsonl` instead of patching the display. I verified every field the parser reads (`thinking`, `toolCall.name/arguments`, `toolName`, `isError`, `model_change.provider/modelId`) against what pi's `session-manager.js` actually appends and the `pi-ai` message types — all match.
3. **Live follow** ✅ — `follow` flag re-arms exactly when scrolled back to bottom (`src/fleet-inspector.ts:387`), matching "follows the live tail until you scroll up".
4. **Non-TUI fallback** ✅ — `buildSnapshotText` path untouched.
5. **Green claim** ✅ reproduced locally (typecheck clean; 544 pass / 0 fail / 3 pre-existing skips in unrelated modules; build OK) and CI shows all 9 runs successful (test win/ubuntu × node 22/24, e2e ×2, pr-validation, build-artifact, docker). The API's `mergeable_state: blocked` is branch protection, not failing checks.
6. **Base** ✅ — head sits directly on current master (`7c23bf5`), fast-forward clean.

## Minor nits (non-blocking, your call)
1. **Claim/code mismatch on list-mode paging**: PgUp/PgDn exist only in transcript mode (`src/fleet-inspector.ts:363-366`); list mode is line-by-line `j/k` only (`:359-361`). Either add `Key.pageUp/pageDown` to the list branch or soften the "both … AND PgUp/PgDn" wording. Low impact since fleet lists are usually short.
2. **Thinking-line right-edge artifact** (`src/fleet-inspector.ts:227,238`): thinking text wraps at `innerW - 2` but the prefix `  ⌥ ` is 4 columns, so full-width wrapped thinking lines exceed `innerW` and `frame()`'s padding truncates them with a trailing `…`. Cosmetic, but visible on dense thinking output. Suggest wrapping thinking at `innerW - 4`.
3. **2 MB head cap** (`src/fleet-inspector.ts:10`): `readHeadSync` reads from file start, so a session exceeding 2 MB stops showing *newest* content (live-follow stales out gracefully at the cap). Edge case; if it matters, switch to tail-reading once the file exceeds the cap.
4. **Hardening**: `UserMessage.content` is typed `string | ContentBlock[]`; `parseSessionJsonl` assumes array. pi 0.83 always writes arrays (checked `agent-session.js`/`messages.js`), so no bug today — a `typeof content === "string"` guard would just immunize against other writers/versions.
5. **Perf nit**: `tick()` re-reads up to 2 MB every 500 ms even when unchanged (the `blocksKey` dedupes parsing, not the read). Negligible on local disk; an `fstatSync` size check before reading would make the idle path free.
6. **Tiny terminals**: `PANEL_MIN_H = 12` can make the overlay taller than a <14-row terminal. Purely cosmetic edge case.

None of these warrant a separate issue per the reporting bar — all fit in this thread. Merge is human-only per AGENTS.md, so the ball is in @ranxianglei's court whenever the nits above are dispositioned (or ignored, they're all optional).

PR #323, opened 2026-09-07
fix(state): keep in-memory subagent compress state across turns (issue #322)

ranxianglei:
Fixes #322

## Problem
In file-less in-memory subagent sessions (pi-subagents `rememberAgents:false` → `SessionManager.inMemory()`, so `getSessionFile() === undefined`), a successful `compress` was silently lost: the next context round reloaded pristine state (`blocks=0`, `nextBlockId=1`) and the model re-compressed the same original context over and over (field report: 152 duplicate compressions, ~$156).

## Root cause
`SessionStateStore.save()` (src/state.ts) hit `if (!file) return;` **before** updating the in-process cache — even though `cacheKey()` already supports file-less keys (`session:<sessionId>`) and `load()` seeds that key with the initial state. Every save in an in-memory session was dropped; every load returned the stale initial entry.

## Fix
Move the cache update ahead of the disk gate in `save()`: cache is updated unconditionally, sidecar persistence still only when a session file exists. One-line reorder in src/state.ts, no API change.

## Tests
- `tests/state.test.ts`: store-level round-trip for a file-less session (save→load keeps blocks + nextBlockId) and isolation between distinct in-memory sessionIds.
- `tests/compress-tool.test.ts`: cross-turn regression — compress m00001 in a file-less session, advance one context round, `acp_status scope:"compressed"` must show `b1 (T1)`; a second compress must be numbered `b2` (guards nextBlockId reset). Verified to FAIL against pre-fix code (negative run done locally).

## Verification
- `npm run typecheck` ✅
- `npm test`: 612 passed, 0 failed, 3 skipped ✅
- `npm run build` ✅ (dist/index.js 719.79 KB, acp-kernel bundled inline)

Boundary noted in the issue thread: `invalidate()` at `session_start` still resets in-memory state across process restarts — out of scope for this fix.

PR #268, opened 2026-09-01
fix: growth scale-flip reset, per-location auto-update, panel denominator

ranxianglei:
Fixes #267 (three related issues).

## 1. Growth baseline mixed two scales (main)

`tokenCount` runs on the estimate scale (chars/4) while the usage anchor is stale, and on the provider scale when it is fresh. A growth delta spanning that switch was a false artifact — e.g. `48k (provider) − 12.5k (estimate) ≈ 35k "growth"` right after a compress, even though the context shrank.

The context transform now tracks the per-session anchor scale (`runtime.noteTokenScale`) and re-anchors the growth baseline (`lastNudgeShownTokens` + `lastPerMessageNudgeTokens`) on a flip, so growth only accumulates same-source deltas. The 0.75/0.95 usage bands keep the floor-stale behavior untouched — PR #258's stale-anchor guard is preserved (`tests/sent-view-arbitration.test.ts` still green).

Note: I did **not** take the "always use `Math.max(sent, providerReal)`" option from the issue, because it would break the stale-anchor guard (the post-compress first nudge would false-trigger emergency from the pre-compress high value). The "growth only accumulates same-source deltas" option is the one that keeps both behaviors correct.

## 2. Auto-updater looped on a read-only global install, failed invisibly

- The check throttle file was shared across every installed copy, so a healthy copy refreshed it and silently suppressed a failing copy's checks; the failing copy then retried the `EACCES` install forever with no user-visible notice.
- The throttle and a new stop-retry marker are now keyed per install location (`locationKey`), so copies don't suppress each other.
- A permission failure (`EACCES`/`EPERM`/…) marks the location read-only and stops retrying, and a one-time hint tells the user to run `npm i -g billion-context-pi` (or remove the global copy).
- Documented the read-only behavior + the two parallel update mechanisms (extension auto-update vs pi core banner) in CONFIGURATION.md / .zh-CN.md.

## 3. Panel percentages used the wrong denominator

`/acp` and `acp_status` measured every percentage against the full window, while the nudge bands run against `(window − output headroom)`. They now apply the same `applyOutputHeadroom` reservation as the live context transform (extracted into a shared helper), so every percentage is measured against the same real request limit.

## Tests

- `tests/growth-scale-flip.test.ts` (new): baseline re-anchors on the stale→not-stale flip; no spurious reset when the scale is stable.
- `tests/update.test.ts`: EACCES → read-only outcome + marker; non-permission failure → no marker; marker present → check skipped + notify once per process.
[...]

PR #328, opened 2026-09-07
fix(#325): floor meter at provider anchor minus tokens reclaimed since it

### LINKED ISSUE #325: [bug] Post-compression sent-view undercounts prompt by ~70–80K → false idle → host-floor emergency re-nudge loop
## Summary

After a successful compression, ACP's own token accounting drops far below the provider-reported prompt size (a roughly constant ~70–80K offset), so the nudge state flips to `idle`. On the next turn the host-floor anchor (#257) snaps `tokenCount` back up to the real prompt size (~88–95% of the effective limit) and re-injects an EMERGENCY "compress now" nudge. The model compresses again, accounting drops again, and the loop repeats — the agent stops working every few turns just to compress.

## Environment

- pi + billion-context-pi v0.1.56
- Model: local vLLM `qwen3.8-27b`, contextWindow=262144, maxTokens=131072
- Every turn logs `[overflow-selfheal] event=output-headroom before=262144 after=131072` → effective limit = 131,072

## Repro session

- Session id: `01a07b3c-ab19-7b19-8290-f7967a8221a6`
- Log: `~/.pi/acp.log` from line 10163 onward
- Compression points: 14:29:55 UTC (blocks b13+b14), 14:51:08 (b15+b16), 14:56:49 (b17)

### 1. Pre-compression estimates match the provider exactly

On nearly every non-compression turn, `[turn] tok == hostTokens`, e.g. @14:27:14 `tok=124726 hostTokens=124726`. The estimator is correct while the conversation is uncompressed.

### 2. Post-compression: constant ~70–80K undercount

| time (UTC) | [turn] tok | provider input+cacheRead | gap |
|---|---|---|---|
| 14:31:30 | 31,745 | 79,788 + 22,528 = 102,316 | 70,571 |
| 14:56:15 | 34,220 | 110,570 (cacheR=0) | 76,350 |
| 14:56:49 | — | 4,153 + 115,200 = 119,353 | — |

The #289 view-recount does not catch this: at 14:56:20 `viewTokens=36,069` vs actual 116,261.

The gap stays in the same ~70–80K band across different compression sizes, which suggests a fixed component is missing from the post-compression sent-view rather than per-block summary miscounting (the 10 active tier-1 block summaries total only ~8–12K tokens).

### 3. Host-floor anchor then re-triggers EMERGENCY

Next turn after each compression, `tokenCount = max(ownEstimate, lastProviderUsage)` snaps back to 106,669 (81%) / 112,658 (86%) / 115,696 (88%) / 118,184 (90%) / 124,879 (95%) of the 131,072 limit → nudge escalates active→emergency and injects:

> ⚠️ Context limit reached — compress now. Prioritize consumed tool outputs.

(acp.log lines 13504-05 @14:27:14, 13848-49 @14:37:21, 13922-23 @14:48:06, 14122 @14:56:20; `voice=emergency emergency=false reInject=false`)

### 4. Loop cycle observed

emergency nudge → model compresses → accounting drops to 21–34K (16–26%, nudge=idle) → 1–2 work turns add tool outputs → host anchor snaps back to ~110–125K (88–95%) → emergency again.

User-visible: agent stopped every few turns to compress instead of working; user interrupted twice (stopReason=aborted @14:52:55, @14:57:08) and typed "继续" twice.

## Secondary symptoms in the looped tail

- Duplicate compress of the identical range m00103..m00149 (second call failed: "Requested range(s) already compressed … Run acp_status")
- One malformed empty `compress {}` call (@14:52:55)
- Model thinking opens with "Need to compress the consumed ranges" on successive turns

These look like a weak local model (qwen3.8-27b) reacting to repeated emergency nudges; the root trigger is the accounting mismatch above.

## Contributing factor

The effective window was halved by output-headroom (maxTokens=131072 reserved out of 262144), putting all nudge bands below the real post-compression prompt floor (~102–135K). Lowering maxTokens to 49152 removes the pressure, but the undercount itself should be fixed.

## Suggested fix direction

Audit what the post-compression sent-view includes (`estimateTokens(coreMessages, coveredIds, imageTokens)` in src/index.ts): something fixed (~70–80K) is present in the actual request but absent from the view only after a compression has occurred. The #289 view-recount compares against the post-processTurn view, so it inherits the same blind spot.

ranxianglei:
Fixes #325.

## Problem
After a successful compress, ACP's own token accounting dropped ~70–80K below the provider-reported prompt size, so the nudge meter flipped to `idle`. On the next turn the host-floor anchor (#257) snapped `tokenCount` back up to the real prompt size (~84–95% of the effective limit) and re-injected an EMERGENCY "compress now" nudge. The model compressed again → loop. The agent stopped working every few turns just to compress.

## Root cause
`src/index.ts` sets `tokenCount = estimateTokens(coreMessages, coveredIds, imageTokens) + systemPromptTokens`, then floors it at the provider-reported prompt size — **unless** `usageAnchorPredatesCompression()` is true (a successful compress landed after the last valid assistant usage), in which case the floor was skipped entirely.

- `estimateTokens` (`src/tokens.ts`) sums message text only. It structurally omits tool-definition schemas, ref-tag overhead, and the injected nudge — a fixed ~70–80K on this host. So the raw sent-view is always that far below the real request.
- Pre-compression, the current-anchor floor lifted `tokenCount` to the provider number — which is why pre-compression estimates "matched the provider exactly."
- Post-compression, the stale-anchor skip dropped `tokenCount` back to the undercounting raw estimate → false `idle`. Next turn, fresh provider usage re-engaged the floor at full real size → spurious EMERGENCY.

The #289 view-recount inherited the same blind spot because it compares against the post-processTurn view.

## Fix
When the anchor predates a successful compress, floor at `(lastProviderUsage − netReclaimedSinceAnchor)` instead of skipping. `netReclaimedSinceAnchor` = Σ over active blocks whose creating `compress` toolResult lands after the last valid usage anchor of `max(0, compressedTokens − summaryTokens)`.

The provider's real number already counts all fixed overhead (tool schemas, etc.) accurately; subtracting only the genuinely-freed tokens yields an exact post-compression size with no need to measure tool schemas — which are unmeasurable/unstable inside the extension (no API enumerates Pi's full tool set). This is why the issue's suggested direction ("add the fixed component to the estimate") was not taken: it would need an unstable magic constant.

New work added after the compress still grows `tokenCount` through `sentTokens` as before; the non-stale path is unchanged; a stale-anchor compress that reclaimed little still correctly trips the nudge.

## Tests
- `tests/floor-stale.test.ts`: unit tests for `compressionAnchorStaleness` (fresh / stale / pre-anchor exclusion / inactive+unattributable skip / negative-savings clamp / multi-block sum).
- `tests/sent-view-arbitration.test.ts`: rewrote the stale test to seed a real reclaiming block (no false emergency), plus a guard test asserting a tiny-reclaim + near-limit context still fires.
- Full suite: 472 tests, 469 pass, 0 fail (3 pre-existing skips). Typecheck + tsup build green.

## Notes
- No version bump (release-branch territory per AGENTS.md).

---

## Rebase onto v0.1.74 (acp-kernel 0.0.81)

Rebased onto current `master`; the single commit is preserved. `src/floor-stale.ts` and `tests/floor-stale.test.ts` applied cleanly (unchanged upstream). Two regions had moved upstream and were merged by hand:

- **`src/index.ts` meter block** — rewritten since this PR branched (#455 provider-calibration cap, #289 sent-view recount). The fix was re-integrated against that structure rather than cherry-picked verbatim.
- **`tests/sent-view-arbitration.test.ts`** — migrated upstream for the acp-kernel 0.0.55 kernel API; the #325 rewrite was reconciled onto it.

### Meter integration
The provider floor is now **continuous** instead of a binary skip:

    hostFloor = max(0, realPrompt − (predates ? netReclaimed : 0))

Fresh turns floor at the full provider number; stale turns floor at `anchor − netReclaimed`. It feeds the same `applyFloors` as the #455 calibration cap and the #289 view-recount, and the #455 calibration stays fresh-only (gated on `!predates`).

### Growth-scale flip signal (#267)
The baseline re-anchor trigger changes from the binary stale flag to `estScaleWins = hostFloor <= sentTokens` — i.e. it fires when the **effective** ruler actually switches between estimate-dominated and provider-floor-dominated. Under the adjusted floor a stale turn can still floor near the full anchor (small reclaim); treating "stale ⇒ estimate scale" would spuriously reset the growth baseline. Keying off which value actually wins avoids that while still catching genuine switches.

### Test adaptations forced by the new base
- **`tests/sent-view-arbitration.test.ts`** — the "no false emergency" case now seeds **120K** reclaimed (was 60K). On acp-kernel 0.0.81 a 60K reclaim leaves the adjusted floor at 63.9% of the window — still inside the ≥45% nudge band, so a nudge correctly fires; 120K → ~30.6% clears the band (the verified idle quadrant). The tiny-reclaim guard test (a genuinely full context must still trip the nudge) is unchanged.
- **`tests/growth-scale-flip.test.ts`** (#267) — the stale→fresh flip case now seeds a real **~150K**-reclaim block so the #325 adjusted floor keeps turn 1 on the estimate scale. Without it, a stale turn with no reclaim reports provider scale — precisely the behavior #325 removes — so the old "stale ⇒ estimate baseline" assumption no longer holds. The stable-scale case is unchanged.

### Verification on the rebased head
`tsc --noEmit` ✅ · tsup build ✅ (acp-kernel inlined, zero runtime deps). All three affected test files pass locally (floor-stale 14, sent-view-arbitration 5, growth-scale-flip 2). No version bump (release-branch territory per AGENTS.md).

### DIFF
--- a/src/floor-stale.ts
+++ b/src/floor-stale.ts
@@ -1,7 +1,7 @@
 // pi's getContextUsage() anchors on the last assistant message with valid
-// provider usage. A successful compress lands AFTER that anchor, so the
-// floor must be skipped for the next LLM call (mirrors pi's own
-// "usage source must be post-compaction" compaction check).
+// provider usage; a successful compress after that anchor leaves it reflecting
+// the pre-compression request until the next turn reports fresh usage.
+import type { CompressionBlock } from "acp-kernel";
 import { isCompressSuccessText } from "./compress-tool.js";
 import { extractText } from "./messages.js";
 
@@ -32,13 +32,16 @@ function validAnchorUsage(u: UsageLike): boolean {
   return total > 0;
 }
 
-/** True when the last valid assistant usage anchor comes strictly BEFORE the
- *  last successful compress toolResult — the host's provider-usage number
- *  still reflects the pre-compression request. Failed/no-op compresses don't
- *  count (nothing was reclaimed). */
-export function usageAnchorPredatesCompression(entries: AnchorEntry[]): boolean {
+interface AnchorScan {
+  lastUsageIdx: number;
+  lastCompressIdx: number;
+  compressIdxByCallId: Map<string, number>;
+}
+
+function scanEntries(entries: AnchorEntry[]): AnchorScan {
   let lastUsageIdx = -1;
   let lastCompressIdx = -1;
+  const compressIdxByCallId = new Map<string, number>();
   for (let i = 0; i < entries.length; i++) {
     const m = entries[i]!.message;
     if (!m) continue;
@@ -52,7 +55,46 @@ export function usageAnchorPredatesCompression(entries: AnchorEntry[]): boolean 
       isCompressSuccessText(extractText(m.content))
     ) {
       lastCompressIdx = i;
+      compressIdxByCallId.set(m.toolCallId, i);
     }
   }
-  return lastCompressIdx > lastUsageIdx;
+  return { lastUsageIdx, lastCompressIdx, compressIdxByCallId };
+}
+
+/** True when the last valid assistant usage anchor comes strictly BEFORE the
+ *  last successful compress toolResult — the host's provider-usage number
+ *  still reflects the pre-compression request. Failed/no-op compresses don't
+ *  count (nothing was reclaimed). */
+export function usageAnchorPredatesCompression(entries: AnchorEntry[]): boolean {
+  const scan = scanEntries(entries);
+  return scan.lastCompressIdx > scan.lastUsageIdx;
+}
+
+export interface AnchorStaleness {
+  // anchor still reflects the pre-compression request
+  predates: boolean;
+  // Σ max(0, compressedTokens − summary) over active blocks whose compress
+  // landed after the anchor; unattributable/pre-anchor blocks are excluded
+  netReclaimed: number;
+}
+
+// issue #325: flooring at a stale anchor's raw value re-fires a false EMERGENCY,
+// while skipping it under-counts by the fixed overhead the host already counts.
+// Return how much was reclaimed since the anchor so callers floor at
+// `anchor − netReclaimed` instead of either extreme.
+export function compressionAnchorStaleness(
+  entries: AnchorEntry[],
+  blocks: readonly CompressionBlock[],
+  countTokens: (text: string) => number,
+): AnchorStaleness {
+  const scan = scanEntries(entries);
+  let netReclaimed = 0;
+  for (const block of blocks) {
+    if (!block.active || block.compressCallId == null) continue;
+    const idx = scan.compressIdxByCallId.get(block.compressCallId);
+    if (idx == null || idx <= scan.lastUsageIdx) continue;
+    const saved = (block.compressedTokens ?? 0) - countTokens(block.summary ?? "");
+    netReclaimed += saved > 0 ? saved : 0;
+  }
+  return { predates: scan.lastCompressIdx > scan.lastUsageIdx, netReclaimed };
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,7 +24,7 @@ import { delegateStatusWidget } from "./fleet-widget.js";
 import { wireToolGuardrails } from "./tool-guardrails.js";
 import { debug, logError, logInfo, logWarn, logThrow, closeLogStream } from "./log.js";
 import { collectCoveredMessageIds, estimateTokens, lastUserMessageId, collectImageTokens, modelSupportsImages } from "./tokens.js";
-import { usageAnchorPredatesCompression } from "./floor-stale.js";
+import { compressionAnchorStaleness } from "./floor-stale.js";
 import { checkForUpdate } from "./update.js";
 import {
   THROTTLE_RETRY_ERROR_MESSAGE,
@@ -244,13 +244,18 @@ function wireContextTransform(pi: ExtensionAPI, runtime: AcpRuntime): void {
       // a heuristic (images, mixed content, per-model tokenizer drift), so
       // the 0.75/0.95 bands run on the real scale via the floor. realUsage is
       // anchored on the last assistant's provider-reported usage + trailing
-      // estimate. Only ever raises (never lowers); skipped while the anchor
-      // predates a successful compress (floor-stale.ts). tokenCount only
-      // feeds processTurn.
+      // estimate. Only ever raises (never lowers). While the anchor predates a
+      // successful compress its raw value still reflects the pre-compression
+      // request, so we subtract the tokens reclaimed since it (issue #325)
+      // rather than skip the floor. tokenCount only feeds processTurn.
       let tokenCount = sentTokens;
       const realPromptTokens = realUsage?.tokens ?? 0;
-      if (!usageAnchorPredatesCompression(entries) && realPromptTokens > tokenCount) {
-        tokenCount = realPromptTokens;
+      if (realPromptTokens > 0) {
+        const { predates, netReclaimed } = compressionAnchorStaleness(entries, state.blocks, defaultCountTokens);
+        const effectiveFloor = predates ? Math.max(0, realPromptTokens - netReclaimed) : realPromptTokens;
+        if (effectiveFloor > tokenCount) {
+          tokenCount = effectiveFloor;
+        }
       }
       // Self-heal (armed): after an overflow, force this turn's usage to >=95%
       // so the kernel's emergency nudge + tool-result truncate fire immediately,
--- a/tests/floor-stale.test.ts
+++ b/tests/floor-stale.test.ts
@@ -1,6 +1,7 @@
 import { test } from "node:test";
 import assert from "node:assert/strict";
-import { usageAnchorPredatesCompression } from "../src/floor-stale.js";
+import type { CompressionBlock } from "acp-kernel";
+import { usageAnchorPredatesCompression, compressionAnchorStaleness } from "../src/floor-stale.js";
 
 // The issue #257 floor must be skipped while the host's provider-usage anchor
 // (the usage on the last valid assistant message) predates the last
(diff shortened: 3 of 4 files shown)


ranxianglei:

[bot] 🏷 Reviewing now — pulling the PR diff, re-verifying the root-cause chain independently, and running the full suite locally before giving a verdict. Will report back shortly.

ranxianglei:

Reviewed end-to-end — **the fix logic is correct** (root cause independently verified, floor math proven in both quadrants), but **the PR cannot land as-is**: it is `dirty` against current master, because master refactored exactly the floor region after your branch point. Details below.

## What I verified
- Pulled head `577a7ac` locally; its own suite: **472 / 469 pass / 0 fail / 3 skip** — matches your claim exactly. Typecheck + tsup green.
- Traced the attribution chain independently: the kernel stamps `compressCallId` onto ranges at apply time and copies it onto the created block (acp-kernel bundle :881 / :2474); the adapter passes the host `toolCallId` through (`src/compress-tool.ts:57,329-330`). `CompressionBlock.active/compressCallId/compressedTokens/summary` exist in both 0.0.48 (your pin) and 0.0.56 (master's pin). The old code already required toolResult entries to carry `toolCallId`, so the new attribution adds no structural assumption.
- Floor math proven end-to-end with instrumented runs: stale anchor 175K − 120K reclaimed → meter floors at **55,006** (30.6% < 45% band → correctly idle, no false EMERGENCY); stale anchor 174K − 1K reclaimed → floors at **173,006** (96.1% → nudge still fires). Both quadrants behave as designed.

## Design verdict
Approving the approach over the alternative: subtracting genuinely-freed tokens from the provider's real number is exact and needs no magic constant; "add the fixed component back to the estimate" would require measuring tool schemas that are unmeasurable/unstable inside the extension. Right call.

## Blocker: dirty merge vs master
Branch point is `cbb753a`; master has since absorbed #289 (sent-view recount), #316/#269 (growth re-inject), #268 (scale-flip reset, panel denominator), #323 (in-memory state save), and release v0.1.59 (acp-kernel 0.0.48 → 0.0.56) — all touching the floor region you modify. Conflicts in `src/index.ts` (imports + floor block) and `tests/sent-view-arbitration.test.ts`.

I resolved it locally and validated: full suite **633 / 630 pass / 0 fail / 3 skip**, typecheck + build green. Recipe for the rebase:

1. Keep master's structure (`armedFloor`, `applyFloors`, the view-recount gate) and replace the stale-skip with:
   `hostFloor = realPromptTokens > 0 ? Math.max(realPromptTokens - (predates ? netReclaimed : 0), 0) : 0`
   `tokenCount = Math.max(sentTokens, hostFloor, armedFloor)`
2. **Semantic decision worth calling out:** the ruler-flip signal passed to `runtime.noteTokenScale` must be `estScaleWins = hostFloor <= sentTokens`, *not* raw `predates`. Under the adjusted floor, a stale anchor with small reclaim still lets the provider floor win (no actual ruler switch); resetting nudge tracking there would zero `lastNudgeShownTokens` so the next nudge passes the growth gate immediately — reintroducing exactly the #325 churn this PR exists to kill. Fresh-quadrant behavior is unchanged except a rare base-wins case that now also resets (more accurate).
3. Test adaptations needed on the rebased tree:
   - `growth-scale-flip`: seed a real 150K-reclaim block before turn 1 — otherwise the no-block stale path floors turn 1 at the full 175K and it is not estimate-scale anymore (its baseline assertion breaks).
   - `sent-view-arbitration` stale case: needs a real reclaiming block plus enough compressible mass outside the protected recent zone (MID pad + ~1K reclaim → 96.1% → fires). With the SMALL pad the whole stream sits inside the protection zone and the kernel legitimately suppresses ("max pending < min benefit 5000") — correct kernel behavior, not a bug.

## Minor finding (report-only, no separate issue per policy)
`/acp`'s statusReport (`src/commands.ts:153,171`) applies the same stale-skip inside a display-only processTurn, so the panel under-reports right after a compress until fresh provider usage arrives. Cosmetic, self-heals next turn, no loop risk — fold into the rebase or take as later cleanup.

**Recommendation:** rebase onto current master with the recipe above (happy to share the exact hunks if useful). Fix code itself: LGTM.

ranxianglei:
**分诊(队列清理 #484): KEEP-VALID(#325 的两极缺陷是真问题),rebase 中等。**

master 现行是二值跳过:`src/floor-stale.ts:36-54` `usageAnchorPredatesCompression()` 在锚点早于成功压缩时整体弃 floor(`src/status-tool.ts:76-81`、`src/index.ts:416/421` 同步跳过)。PR 的 `compressionAnchorStaleness()` 返回 `netReclaimed`(floor=anchor−netReclaimed)在 master 无对应物(grep `netReclaimed|reclaim` 仅内核内部)。恰好治"陈旧 floor 假紧急 vs 跳过又漏计"两个极端。内核依赖(CompressionBlock.compressCallId/compressedTokens)0.0.81 均在。

**rebase 风险:中等。** `floor-stale.ts` hunk 漂移小,但 `src/index.ts` meter 块被 #455/#289 重写——`anchor−netReclaimed` floor 与 `measured×1.2` 校准上限的交互必须在 rebase 里显式定义;`tests/floor-stale.test.ts` 与 `tests/sent-view-arbitration.test.ts` 两个现存测试文件需合并。

ranxianglei:

已按分诊完成 rebase 到当前 master(v0.1.74 / acp-kernel 0.0.81),单 commit 保留。head 现为 `55ca193`。

**冲突解决**:`src/floor-stale.ts`、`tests/floor-stale.test.ts` 干净套用(master 未动);`src/index.ts` meter 块(被 #455/#289 重写)与 `tests/sent-view-arbitration.test.ts`(被 acp-kernel 0.0.55 迁移改动)手工合并。

**关键交互(分诊点名的"floor 与校准上限的交互")**:provider floor 从二值跳过改为连续值
`hostFloor = max(0, realPrompt − (predates ? netReclaimed : 0))`,接入 #455 校准上限与 #289 view-recount 的同一 `applyFloors`;#455 校准保持 fresh-only(gate on `!predates`)。另把 #267 growth-baseline 的 flip 触发从二值 stale 标志改为 `estScaleWins = hostFloor <= sentTokens`(以实际生效的 ruler 为准)——否则"陈旧但少回收、仍锚在近满额"的轮次会被误判为换尺而重置增长基线。

**测试适配(新基座强制,已在 PR 描述补记)**:
- `sent-view-arbitration`:「无假紧急」用例回收量 60K→**120K**。0.0.81 上 60K 回收后 adjusted floor = 175K−60K = 63.9%,仍落在 ≥45% nudge 带内会正确触发;120K→~30.6% 才清出(即前面楼层验证过的 idle 象限)。tiny-reclaim guard 用例不变。
- `growth-scale-flip`(#267):stale→fresh 用例补种一个真实 **~150K** 回收块。否则「陈旧且零回收」会落到 provider 尺度(正是 #325 要消除的行为),原「stale⇒estimate baseline」假设不再成立。stable 用例不变。

**验证**:typecheck ✅ / build ✅(acp-kernel 内联,零运行时依赖)/ 三个受影响测试文件全绿(floor-stale 14、sent-view 5、growth-scale 2)。本地全量套件里非绿的仅为环境噪声:临时文件写入受限导致的 EACCES(CI 有可写 /tmp)+ 2 个 `delegate-entry` 主机解析断言(该文件只 import `delegate-tool.js`,与 master 字节一致,在 master 上同样失败,与本 diff 无关)。**无任何失败与本 PR 改动相关。**

未改 `status-tool.ts` / `commands.ts` 的展示层 stale-skip(纯显示、自愈,前面楼层已记录),留作后续。CI 在新 commit 上运行中。

PR: https://github.com/ranxianglei/billion-context-pi/pull/328

ranxianglei:

补充:rebase 后的 head `55ca193` CI 已全部通过 —— **PR Build Artifact ✅ / CI ✅ / E2E ✅**(干净环境下全量套件通过,本地那批 EACCES / delegate-entry 环境噪声在 CI 不复现)。PR 现为 mergeable,待人工合并。

一句话:已把 #328 rebase 到当前 master 并解决 meter 块与两处测试的冲突(连续 floor + estScaleWins flip),CI 全绿,可合并。

PR #337, opened 2026-09-08
report compressed ranges in the success panel (#335)

### LINKED ISSUE #335: 压缩成功也报告压缩范围,防止模型重复压缩同一个范围
压缩成功也报告压缩范围,防止模型重复压缩同一个范围

ranxianglei:
Fixes #335

## Problem

The `compress` success panel only reported token stats:

```
▣ ACP | 58.5K → 5.7K tokens (~52.8K reclaimed, 4 blocks)
```

The model never learned *which* refs were consumed, so it could re-issue the same (or overlapping) ranges. The kernel rejects those as "already compressed" no-ops, but the model has no signal in the success path telling it the range is gone.

## Fix

When `blocksCreated > 0`, the panel now appends the consumed ranges (the kernel already records `startRef`/`endRef` on every new block, and the tool passes the model's refs through verbatim at `src/compress-tool.ts:350`):

```
▣ ACP | 58.5K → 5.7K tokens (~52.8K reclaimed, 4 blocks)
Compressed: b3 (m00010..m00045), b4 (m00046..m00060) — these refs are now consumed; do not compress them again.
```

- Line 1 is byte-identical to before; panel parsers (`compressPanelBlocks`, `isCompressSuccessText`, `isCompressNoopText`) only read line 1, so no behavior change for floor-stale / replay classification.
- No-op runs (`blocksCreated === 0`) and rejection panels are unchanged.
- Batch calls list every created block; partial failures still show `Errors:` for the rejected ranges.

## Tests

New regression test `tests/compress-tool.test.ts` — "compress success panel reports the compressed ranges" — asserts the panel contains `Compressed: b1 (m00001..m00002)` after a successful compress.

`npm run typecheck && npm test && npm run build` — all green (624 pass, 0 fail).

### DIFF
--- a/src/compress-tool.ts
+++ b/src/compress-tool.ts
@@ -431,6 +431,9 @@ async function handleCompress(args: CompressArgs, runtime: AcpRuntime, ctx: Exte
   }
 
   const lines = [`▣ ACP | ${formatK(beforeTokens)} → ${formatK(afterTokens)} tokens (~${formatK(reclaimed)} reclaimed, ${blocksCreated} block${blocksCreated > 1 ? "s" : ""})`];
+  if (blocksCreated > 0) {
+    lines.push(`Compressed: ${newBlocks.map((b) => `${b.blockId} (${b.startRef ?? "?"}..${b.endRef ?? "?"})`).join(", ")} — these refs are now consumed; do not compress them again.`);
+  }
   if (warnings.length > 0) lines.push("⚠️ " + warnings.join("; "));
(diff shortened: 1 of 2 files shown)


ranxianglei:
**已被 #377 取代**:同为『压缩成功面板报告消费范围』,#377 已合并且实现更完整(blockSpanLabel + 空洞 `*` 标记 + Tn 后缀 + kernel formatCreatedBlocks)。关闭本 PR;#335 由 #377 关闭闭环。

Issue #348, opened 2026-09-09
reasoning-drop never fires in single-user-message marathon sessions (closed-turn gate unreachable)


ranxianglei:
Session 01a086ff-92ff-76a5-a141-3c839666877d (/tmp, qwen3.8-27b, v0.1.62-pr.347.141, ~/.pi/acp.json compress.reasoning = {drop:true, threshold:1}): 64 entries, genuine user text messages ONLY at lines 3/5 (the opening prompt); 3 closed compress calls (lines 34/49/60) each carrying 20612 / 8411 / 10581 chars of thinking on the compress-call assistant message; zero [reasoning-drop] events — the ~39.6k-char thinking floor rode EVERY request.

Instrumented dist (patched call site): cfg={drop:true,threshold:1} arrives correctly; lastUserIdx=1 (the OPENING prompt); compressMsgs=30 — 30 >= 1, outside the closed zone.

Root cause: dropCompressReasoning (src/reasoning-drop.ts, #339; same gate in billion-context #651/PR #667) treats messages strictly before the last genuine user message as closed. In a single-prompt agent marathon — one user message at the head, model works the whole session, exactly the scenario that builds the largest thinking floor — lastUser sits at the session head, so every compress turn is forever the active round and is never dropped.

Proposed fix: close on round evidence — a compress call is closed once its tool result has arrived AND any later message exists (the conversation moved past it). Provider safety stays via per-provider layering (drop:false for reasoning-echo providers, e.g. GLM reasoning_content).

Cross-refs: pi #339 (merged), billion-context #651 / PR #667 (open, same gate proxy-side).

PR #363, opened 2026-09-10
fix(reasoning-drop): auto-disable compress reasoning drop on strict-echo upstreams (#361)

### LINKED ISSUE #361: Twin of billion-context#684: reasoning-drop needs a strict-echo gate (DeepSeek thinking mode)
The adapter's #651 twin (src/reasoning-drop.ts, drop:true default) strips the reasoning run preceding a closed compress call — on strict-echo thinking upstreams (DeepSeek: `reasoning_content must be passed back` when tools are in play) that produces the same HTTP 400 fixed in billion-context#690 (proxy side) + acp-kernel#245 (fold atomicity).

The kernel-side fix protects the fold split for every host, but the pi-side DROP itself has no strictness gate. Suggested: port `isStrictReasoningEcho()` (static deepseek host check from the provider baseUrl + learned flag), or at minimum document `compress.reasoning.drop:false` for DeepSeek/GLM-thinking/QwQ provider configs.

Depends on acp-kernel 0.0.63 (turn-integrity) for the full closure; pin bump will come with the next release PR.

ranxianglei:
## What

Auto-disable the `compress.reasoning.drop` pass on strict-echo thinking upstreams so a DeepSeek thinking-mode session can't be broken by the default `drop:true`.

## Why

`dropCompressReasoning()` (default `drop:true`) strips the reasoning run preceding a closed compress round. On strict-echo thinking upstreams (DeepSeek thinking mode: `reasoning_content must be passed back` when tools are in play) that **independently** reproduces the HTTP 400 on the *next* request — the same failure class fixed proxy-side in [billion-context#690](https://github.com/ranxianglei/billion-context/pull/690) and kernel-side (fold atomicity) in [acp-kernel#245](https://github.com/ranxianglei/acp-kernel/issues/245).

Verified against the original report in billion-context#684, which states the closed-compress DROP is a loss path distinct from the fold split and reproduces the identical `invalid_request_error` ("The `reasoning_content` in the thinking mode must be passed back to the API") even after a single closed compress round.

#348 tightened closed-round detection and kept a **manual** per-provider `drop:false` escape hatch, but the drop pass itself had no provider strictness gate — this PR adds the auto gate.

## How

Ported the sibling `isStrictReasoningEcho()` detector, adapted to pi's signal sources:

- New pure `isStrictReasoningEcho(provider?, baseUrl?)` — static DeepSeek check from `ctx.model.baseUrl` / provider name (case-insensitive substring).
- New pure `applyStrictReasoningGate(cfg, provider?, baseUrl?)` — forces `drop:false` on a strict-echo upstream, preserving `threshold`; no-op (same reference) otherwise.
- Wired into `runtime.reasoningDropFor(ctx)`; logs `compress-reasoning-auto-disabled` once per session.

The gate overrides an explicit `drop:true` (safety wins). Cost-free for non-thinking models on those hosts (they emit no `thinking` parts). GLM-thinking / QwQ / self-hosted mirrors are deliberately **not** auto-detected (avoids disabling the pass for their non-thinking models) — they keep the documented manual override.

No learn-on-400 self-heal here: the in-process adapter does not own the HTTP layer (Pi makes the LLM call; no extension hook observes the upstream 400), unlike the proxy.

## Scope

- 6 files: `src/reasoning-drop.ts`, `src/runtime.ts`, `tests/reasoning-drop.test.ts`, `CHANGELOG.md`, `CONFIGURATION.md`, `CONFIGURATION.zh-CN.md`.
- Does **not** bump the `acp-kernel` pin (stays at master's 0.0.62). Full closure pairs with acp-kernel 0.0.63 turn-integrity (#245), which lands in the next release PR.
- Verified locally: typecheck clean, 700 tests (697 pass / 0 fail / 3 skipped), build success.

中文摘要：为 DeepSeek 思考模式等严格回传上游自动禁用 compress reasoning drop（默认 drop:true 会丢 reasoning 触发 HTTP 400），移植代理侧 #690 的静态 deepseek 检测并接入 runtime，GLM/QwQ/自托管镜像仍走手动 drop:false；已验证 typecheck/测试/构建全绿。

Fixes #361


ranxianglei:

[bot] 🏷 收到,开始 review PR #363(auto-disable compress reasoning drop on strict-echo upstreams)。先做重复筛查 + 合理性审核(问题复现/根因层次/方案评估),再查 diff 整洁度,随后本地跑 typecheck / 700 tests / build。有结论会回贴。

ranxianglei:

**Review complete — approved, ready to merge.**

**Triage (bug class — verified before touching anything):**
- The reported failure is real and this hits the right layer. `dropCompressReasoning()` (`src/reasoning-drop.ts:133`) strips `thinking` parts from closed-round compress messages under the default `drop:true` (threshold 2048) with no provider awareness — on DeepSeek thinking mode the next request loses `reasoning_content` → HTTP 400. This is a loss path **distinct** from the kernel fold split ([acp-kernel#245](https://github.com/ranxianglei/acp-kernel/issues/245)) and the proxy fix (billion-context#690): neither covers this in-process pass, so gating it pi-side is the correct place, not a band-aid.
- Signal source verified against the installed host types: `baseUrl` and `provider` are real fields on pi-ai's `Model<TApi>` (`dist/types.d.ts:647`) and `ExtensionContext.model` (pi-coding-agent 0.83.0), so detection works at runtime — not just plausible typing.
- Design checks out: safety-first (explicit `drop:true` loses), `threshold` preserved, same-reference no-op for non-strict upstreams, log-once per session, GLM-thinking/QwQ/self-hosted mirrors deliberately left on the manual override. Skipping learn-on-400 self-heal is right here — the in-process adapter doesn't own the HTTP layer.
- Dup screen: extends #336/#348 rather than duplicating them; twin fixes live in other repos.

**Verification (run locally on the branch):** typecheck clean · 702 tests, 0 fail / 3 skipped · build success.

**Two small issues I fixed directly on the branch** (head is now `5c70710`; also rebased onto current master `a8dfb49` since the PR was behind — only `package.json`/lockfile moved, conflict-free):
1. `test(reasoning-drop)` — added `tests/e2e-reasoning-drop-gate.test.ts` (+2 e2e tests). The pure-function tests alone would still pass if someone removed the gate call in `runtime.reasoningDropFor` or stopped passing `ctx.model.baseUrl` into it; the new file drives the real `createRuntime → reloadConfig → reasoningDropFor` pipeline and covers: baseUrl-only detection (provider name without "deepseek"), explicit `drop:true` losing to the gate, threshold survival, and exactly-once-per-session logging.
2. `docs(configuration)` — pre-existing render bug in the exact paragraph this PR documents: `CONFIGURATION.md:536` contained a literal `\n` gluing the ` ```json ` opener to the prose line, so the fence never opened and the next physical fence line started a stray code block that swallowed the following sections in rendered output. Split onto its own line (matches `CONFIGURATION.zh-CN.md`). Revert-able as its own commit if you disagree.

The json-example re-indent in your diff is fine — it aligns the fence with the surrounding block's 3-space convention.

Full closure pairs with acp-kernel 0.0.63 turn-integrity (#245) in the next release PR as you noted — nothing further needed here.

中文一句话：审核确认 strict-echo 自动禁用 reasoning drop 的修复定位正确（pi 侧 DROP pass 缺严格回传门控，属独立丢失路径，与内核/代理修复互补），已直接补齐 e2e 接线测试并修复 CONFIGURATION.md 一处既有代码围栏渲染 bug，rebase 至最新 master 后 typecheck/702 测试/构建全绿，可以合并。

Issue #271, opened 2026-09-01
Feature parity: port the upstream `export` command (session handoff export)


tjp72:
The upstream [billion-context](https://github.com/nickthecook/billion-context) ships a session `export` command (`bili export`) built on acp-kernel's `prune()` renderer. The Pi adapter (`billion-context-pi`) does not port this capability — sessions compressed inside Pi cannot be exported as a readable handoff document. This issue tracks restoring feature parity.

**What upstream has**

`src/export.ts` (~175 lines) plus a CLI command:

```
bili export                              # list persisted sessions (id/title/label/tokens/blocks)
bili export <session-id|label>           # render handoff markdown
bili export <id> --output handoff.md     # write to file
bili export <id> --full                  # full original messages instead of folded view
```

Key behaviors worth preserving:

1. **Folded view by default** — uses acp-kernel's `prune(messages, state)` to render the conversation exactly as the model saw it: compressed ranges are replaced by their block summaries (`[Compressed conversation section]` placeholders), never the raw messages.
2. **`--full` mode** — renders every original message: from the v3 conversation snapshot when available, otherwise falls back to the `blockContents` cache (v2 sessions).
3. **v2 fallback** — sessions without a persisted snapshot still export block summaries (with originals only under `--full`), plus an explicit note when nothing is recoverable.
4. **Handoff semantics** — the rendered document ends with guidance to paste the block summaries into a new session to continue without the proxy.

**Why this matters for the Pi adapter**

- Sessions managed by the adapter are currently opaque after the fact — there is no way to produce a human-readable, fold-aware snapshot of a compressed conversation.
- The handoff document is the only mechanism to migrate context between different hosts or to a fresh session without the proxy.
- `prune()` already lives in `acp-kernel` (which Pi pins and bundles); only the consumer code (`export.ts` + CLI wiring) is missing.

**Proposed scope**

1. Port `export.ts` (listSessions / exportSession / renderHandoff) to the Pi source tree.
2. Wire the `export` command into the Pi CLI (or expose it as a tool/command if Pi has no CLI surface — adapter-appropriate integration).
3. Keep both output modes (`--output` to file and stdout return) and the `--full` flag.
4. Preserve the v2 fallback path so older persisted sessions remain exportable.

**Acceptance criteria**

- [ ] `export` with no selector lists persisted sessions with title/label/tokens/blocks.
- [ ] `export <id>` renders the folded view (block summaries in place of compressed ranges, in conversation order).
[...]

Issue #421, opened 2026-09-13
toolOutputMaxBytes default 200KB is too loose for the modern tool path


ranxianglei:
## Problem

`toolOutputMaxBytes` defaults to 200000 while pi core caps bash/read/grep results at 50KB (bash: 50KB→full output to temp file). The 200KB net therefore only ever engages for tools pi does not cap (MCP/custom), and when it does engage it lets a 200KB blob into context — 4x the core ceiling, instantly eating ~4% of a 50K window (see #398 small-context work) or triggering a forced compress.

## Proposed fix

Default `50_000` — aligned with pi core's own caps so every tool path lands under one ceiling. Still overridable per acp.json (`toolOutputMaxBytes`), `0` disables.

## Notes
- No test asserts the old default; guardrail tests pass explicit caps.
- Docs (CONFIGURATION.md + zh-CN) updated in the same PR.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework-aio) · qwen3.8-27b

[bot] 🏷 Triage done, implementing now — verification summary before the PR:

**Duplicate screening:** no duplicate found. Closest relative is #210 (the wiring fix that made the documented default actually reachable) — that fixed *whether* the default applied; this issue is about *what* the default value is. Distinct, proceeding.

**Verified claims:**
- ✅ `DEFAULT_TOOL_OUTPUT_MAX_BYTES = 200_000` at `src/config.ts:270`, wired via `?? DEFAULT_TOOL_OUTPUT_MAX_BYTES` at `src/tool-guardrails.ts:241`.
- ✅ The "pi core caps bash/read/grep at 50KB" premise is corroborated by the repo's own documentation — the field's doc comment (`src/config.ts:216-217`: "Pi already caps bash/read/grep at 50KB/2000 lines (bash full output is saved to a temp file)") and the #210 changelog entry.
- ⚠️ One nuance vs. the issue body: nothing pins the literal `200000` in a test, but `tests/tool-guardrails.test.ts:166` has a test *named* "applies the documented 200KB default when toolOutputMaxBytes is unset" with a loose bound (`< 250 * 1024`). It passes under the new default too, but I'll rename it and tighten the bound to `< 60 * 1024` so the default contract is actually asserted.

[...]

PR #414, opened 2026-09-12
fix(delegate): configurable fleet-inspector shortcut + move default off ctrl+alt+f (#412)

ranxianglei:
## Problem

With both `billion-context-pi` (ACP) and `pi-subagents` installed, Pi prints an orange warning every startup:

> Extension shortcut conflict: 'ctrl+alt+f' registered by both billion-context-pi and pi-subagents. Using pi-subagents.

Both extensions unconditionally hard-code `ctrl+alt+f`. Pi's loader (`getShortcuts()`) only warns on cross-extension conflicts and lets the later-loaded binding silently win — no dedup, no user override. So one of the two fleet inspectors can't be opened from the keyboard.

Same conflict reported on the other side: nicobailon/pi-subagents#2196

## Fix

Make ACP's fleet-inspector key configurable **and** move its default off the collision:

- New `acp.json` option `delegate.fleetShortcut` (any Pi key id); empty string `""` disables registration.
- Default `ctrl+alt+f` → `ctrl+alt+d`: the inspector is already reachable via `/acp-fleet`, while pi-subagents' status bar advertises `Ctrl+Alt+F Fleet` (muscle memory lives there), so ACP yields. Matches the reporter's local workaround.
- TUI widget hint now renders the effective key dynamically (`formatShortcutLabel`, mirroring pi-subagents casing) instead of a hard-coded string.
- Non-string values fall back to the default with a logged warning (never fail startup).

Files: `src/config.ts` (`DEFAULT_FLEET_SHORTCUT` / `resolveFleetShortcut` / `DelegatePolicy`), `src/index.ts` (register per policy at session_start), `src/fleet-widget.ts` (dynamic hint), `tests/fleet-shortcut.test.ts` (new), `CONFIGURATION.md` (+zh-CN), `CHANGELOG.md`.

## Verification

- `npm run typecheck` → clean
- `npm test` → 703 tests, 700 pass, 0 fail, 3 skipped (incl. 7 new)
- `npm run build` → clean; `@earendil-works/pi-tui` import is type-only (erased, no runtime dep)

## One decision point

The new default `ctrl+alt+d` resolves the collision without touching pi-subagents and is fully overridable via `delegate.fleetShortcut` (or `""` to disable). If you'd prefer ACP keeps `ctrl+alt+f` and pi-subagents yields instead, say so and I'll flip it.

Fixes #412

PR #425, opened 2026-09-13
fix(prompts): rule-slot sections accept string overrides (#423)

### LINKED ISSUE #423: sanitizePromptSections drops rule-slot STRING overrides — kernel master leanPack carries a condensed howToCompress text that will be silently discarded
## Problem

Discovered while validating a local kernel-integration build (kernel master @ c843f48, after PR #269): kernel leanPack's `adapters.pi.promptSections.howToCompress` changed from `null` to a full condensed `HOW TO COMPRESS (condensed)` string (kernel src/packs.ts:197). npm acp-kernel 0.0.66 still has `null`, but the next kernel release will ship the string.

pi's `sanitizePromptSections` (src/system-prompt.ts:82) only accepts:
- `string | null` for PROMPT_SECTION_KEYS (acpTags, summariesInContext, tools, whenToCompress, ...)
- `null` ONLY for RULE_SLOT_KEYS (philosophy, howToCompress, tier2, tier3)

So a string-valued rule-slot override is silently dropped → lean's condensed how-to-compress rules vanish on pi when kernel ships it (lean falls back to the default howToCompress section — silent behavior regression), and `tests/prompt-pack.test.ts`'s `piAdapterSurface(leanPack)` assertion (`howToCompress` === null) will fail against the newer kernel.

## Proposed fix
- Decide the contract: rule slots accept `string` (replace) as well as `null` (remove) in both sanitize and `buildAcpSystemPrompt`'s section application — OR kernel keeps rule slots null-only and ships condensed rules via `prompts` overrides instead
- Add pi-side test with a string rule-slot override; keep CI green on 0.0.66 and on the next kernel bump

## Environment
- pi master @ 975cc43 (post #403 migration), acp-kernel pinned 0.0.66
- kernel master @ c843f48 already carries the change

ranxianglei:
Fixes #423

## Problem
`sanitizePromptSections` accepted `null` only for the rule slots (philosophy / howToCompress / tier2 / tier3). Kernel master (post-#269, formalized by open PRs #263/#266) ships lean's `howToCompress` as a **condensed string** (~2.5K, restoring the summary style floor lost in the #263 hallucination incident). On the next kernel bump pi would have silently dropped lean's condensed rules (falling back to the 4.5K default — silent regression) and the #403 leanPack assertion would have gone red.

## Fix
- Rule slots now accept `string | null` in sanitize (string = replace slot body, null = remove) — the apply path (`applySectionOverrides`, kernel) already treats strings as replacement, so this is a one-line contract fix
- Tests: string override survives sanitize + replaces the slot text in `buildAcpSystemPrompt`; kernel-master-style condensed howToCompress flows through `piAdapterSurface`; junk test updated to the new contract (strings kept, non-strings still dropped)

## Verification
805 tests, 0 fail (3 pre-existing skips), typecheck clean, build green. Also validated against a local kernel-integration build (master + #271 + #273): the condensed string flows into the rendered prompt.


ranxianglei:

[bot] 🏷 Review started. Initial finding: the branch carries **two** commits — the #423 rule-slot fix (`bbfcc60`) **and** an unrelated guardrails change (`7deef59`, refs #421, config.ts + CONFIGURATION*.md). That makes the diff off-scope for this PR; I'll verify the code, run the checks, then converge the branch so every line is #423-related (the #421 change will get its own PR so nothing is lost). Full review reply shortly.

ranxianglei:

[bot] 🏷 **Review complete — diff converged, one small fix added directly to the branch, ready to merge.**

## Duplicate screening
No duplicate of the #423 fix exists. But the branch carried a **second, unrelated commit** (`7deef59`, "feat(guardrails): default tool output cap 200KB -> 50KB", refs #421) touching `src/config.ts` + both CONFIGURATION docs. That change **already has its own PR**: [#426](https://github.com/ranxianglei/billion-context-pi/pull/426) (open, branch `2026-09-13_tool-output-default-50kb`) — and #426 is the *more complete* version (same constant change plus the `tool-guardrails.ts` wiring comment, the renamed/tightened default-contract test, and a CHANGELOG entry; the bundled commit had none of those). Verdict: duplicate with a superior existing counterpart → dropped from this branch, nothing lost.

## Verification of the fix itself
- **Kernel contract claim holds.** Pinned acp-kernel 0.0.66, `applySectionOverrides` (`node_modules/acp-kernel/dist/index.js:806-814`): `o === null` → section removed; `typeof o === "string"` → replacement pushed; otherwise default kept. So yes — the apply path already treats strings as replacement; the sanitize side was the only broken half of the tri-state contract. One-line contract fix, correctly scoped.
- **Zero behavior change at the current pin.** Inspected the pinned kernel's lean pack pi surface: every slot except `acpTags` is `null` today (`howToCompress = null`). This is purely forward-compat for the next kernel bump that ships the condensed ~2.5K string — exactly as described. No silent regression introduced either way.
- **Checks re-run locally on the branch:** `npm run typecheck` clean · `npm test` **805 tests / 802 pass / 0 fail / 3 skipped** — the same 3 pre-existing skips as master (I cross-checked skip names on both branches; an initial run showed 6 skips but that did not reproduce — transient, not PR-related) · `npm run build` green (794.68 KB dist).

## What I changed on the branch (committed directly, per review policy)
1. **Dropped `7deef59`** via `git rebase --onto origin/master` — see duplicate finding above.
2. **Fixed a type inconsistency the fix left behind** (`3bf4300`): `src/system-prompt.ts:14-17` still declared the four rule slots as `philosophy?: null; howToCompress?: null; tier2?: null; tier3?: null;` while `sanitizePromptSections` now stores replacement strings there (via cast). The interface lied about the new capability — typed consumers would see these fields as never-string. Widened all four to `SectionOverride` (`string | null`), matching the kernel tri-state contract and making the interface uniform. No consumer depends on the old null-only typing (checked all 6 usage sites: config/surface/user-config/prompt-pack/index). Re-ran typecheck + full suite + build after the edit: all green.

**Branch state after convergence** (force-pushed with `--force-with-lease`, CI will republish the `pr-425` artifact):
```
3bf4300 fix(prompts): type rule-slot sections as SectionOverride … (refs #423)   ← reviewer fix
330306f fix(prompts): rule-slot sections accept string overrides … (refs #423)   ← original fix
```
Diff vs master is now exactly `src/system-prompt.ts` (+10/−4) and `tests/prompt-pack.test.ts` (+31/−2) — every line is #423-related. New tests are meaningful: string survives sanitize *and* replaces the rendered slot text (with a guard that the 4.5K default text is gone), the kernel-master-style condensed string flows through `piAdapterSurface`, and the junk test pins the new contract (strings kept, non-strings dropped).

## Minor observations (no action needed)
- An empty-string override now replaces a slot with `""` (renders as a blank gap in the `\n\n` join) rather than being dropped — consistent with kernel semantics for every other section, harmless.
- PR body says "3 pre-existing skips" — matches my verified count.

中文摘要：修复本身正确且已验证（kernel 的 `applySectionOverrides` 本就是 string=替换/null=删除 三态契约，当前 pin 下行为零变化，纯为下个 kernel 版本的前向兼容）；发现分支里混入了与 #421 相关的无关提交，而该改动已有更完整的独立 PR #426 覆盖，故已从本分支剔除（无工作丢失）；另直接在分支上补了一个小修复——`PiPromptSections` 四个规则槽的类型仍声明为 `null`-only、与新契约不符，已改为 `SectionOverride`。diff 现已收敛到与本 PR 目的完全相关，typecheck/805 测试/build 全绿，可以合并。

ranxianglei:
Closing as superseded by #427 — same sanitize contract fix (rule slots accept string/null), #427 adds CONFIGURATION docs + surface-config tests and is based on newer master (post-#414).

PR #422, opened 2026-09-13
feat(guardrails): default toolOutputMaxBytes 200KB -> 50KB

ranxianglei:
Fixes #421

- `DEFAULT_TOOL_OUTPUT_MAX_BYTES = 50_000` — matches pi core's bash/read/grep caps so every tool path lands under one ceiling; the cap is the last-resort net for tools pi does not cap itself (MCP/custom tools)
- Still fully overridable (`toolOutputMaxBytes` in acp.json, `0` = disable); no behavior change for existing configs that set it explicitly
- CONFIGURATION.md + zh-CN updated (example, quick-reference table, section body)
- Suite: 803 tests, 0 fail (3 pre-existing skips), typecheck clean, build green

PR #426, opened 2026-09-13
fix(guardrail): toolOutputMaxBytes default 200KB → 50KB, aligned with pi core cap

### LINKED ISSUE #421: toolOutputMaxBytes default 200KB is too loose for the modern tool path

ranxianglei:
Fixes #421

## Problem
`toolOutputMaxBytes` defaulted to 200000 while pi core caps bash/read/grep at 50KB (bash full output saved to a temp file). The 200KB net therefore only ever engaged for tools pi does not cap (MCP/custom) — and when it did, it let a 200KB blob (~4× the core ceiling) into context: several percent of a small window at once, or a forced compress (see #398 small-context work).

## Change
- `src/config.ts`: `DEFAULT_TOOL_OUTPUT_MAX_BYTES = 50_000` + doc comment updated (alignment rationale; "set higher for large MCP outputs" escape hatch documented)
- `src/tool-guardrails.ts`: wiring comment 200KB → 50KB (code unchanged — it reads the constant)
- `tests/tool-guardrails.test.ts`: default-contract test renamed 200KB→50KB and bound tightened `< 250*1024` → `< 60*1024`, so the default is actually asserted (previously any value under 250KB passed)
- `CONFIGURATION.md` + `CONFIGURATION.zh-CN.md`: example config, summary table, detailed section (default + description)
- `CHANGELOG.md`: Unreleased entry

Semantics preserved: explicit acp.json override wins; `0`/negative still disables.

## Verification
- `npm run typecheck` clean
- `npm test`: 693 pass / 0 fail / 3 pre-existing skips
- `npm run build` OK


ranxianglei:
Rebased onto current master (CHANGELOG conflict resolved — both entries kept, #421 entry on top). 803 tests / 0 fail / typecheck clean against pin 0.0.66. Supersedes #422 (closed).

PR #427, opened 2026-09-13
fix(surface): accept string overrides on prompt-section rule slots

### LINKED ISSUE #423: sanitizePromptSections drops rule-slot STRING overrides — kernel master leanPack carries a condensed howToCompress text that will be silently discarded

ranxianglei:
## What

`promptSections` rule slots (`philosophy`/`howToCompress`/`tier2`/`tier3`) now take the same tri-state as the nine structural sections: a `string` **replaces** the section, `null` **removes** it. Previously they were null-only and a string override was silently dropped by `sanitizePromptSections`.

## Why

Kernel master (acp-kernel#269, c843f48) changed `leanPack.surface.adapters.pi.promptSections.howToCompress` from `null` to the condensed `HOW TO COMPRESS (condensed)` string (kernel src/packs.ts:197). Under pi's old null-only contract, that release would (a) silently discard lean's condensed how-to-compress rules on pi and (b) break `tests/prompt-pack.test.ts`'s `howToCompress === null` assertion. The `adapters.pi.*` contract is owned by pi (the kernel treats host adapters as opaque), so pi adapts.

Contract decision (discussed in #423): accept strings on rule slots rather than rework the already-merged kernel work to ship via `surface.prompts` — the latter would route lean's builtin rules through the `acknowledgePromptsRisk` gate, dropping them for users who didn't set the flag while keeping every other lean trim.

## Changes

- `src/system-prompt.ts`: widen `PiPromptSections` rule-slot fields to `SectionOverride`; unify the sanitize condition across all 13 keys. Application path untouched — `applySectionOverrides` already implements string=replace / null=remove, and sections apply after `prompts`, so a string section override wins over a `prompts` override.
- `tests/surface-config.test.ts`: new test — string replaces a rule slot in the built prompt, null removes it, junk types and unknown keys still dropped.
- `tests/prompt-pack.test.ts`: lean assertions are version-tolerant (sanitized value must equal whatever the kernel ships — null on 0.0.66, string on master); end-to-end assertion that the condensed text reaches the built prompt when the kernel ships it.
- `CONFIGURATION.md` / `CONFIGURATION.zh-CN.md`: document all 13 keys including the four rule blocks.

## Verification

- Pinned acp-kernel 0.0.66: `npm run typecheck && npm test && npm run build` → 803 tests, 0 failures, build OK.
- Local kernel-master build @ c843f48 overlaid onto node_modules (local only, not committed): full suite green (803 tests, 0 failures), including the new "condensed rules reach the prompt" assertion.
- No `version` bump, no lockfile changes.

Fixes #423


ranxianglei:

[bot] 🏷 **Reviewed — LGTM, ready to merge.**

**Duplicate screening:** no overlapping open issue/PR found for this change; #427 is the linked fix for #423 (the originating contract-drift report), so no duplicate action needed.

**Diff cleanliness:** 5 files / +29 −15, every line on-topic. No lockfile changes, no `version` bump, no stray reformatting. Clean.

**Contract verification (independent of the author's claims):**
- Read the kernel's `applySectionOverrides` in the pinned acp-kernel 0.0.66 dist (`dist/index.js:806-814`): it is key-agnostic — `null` skips the section, `string` replaces it, anything else keeps the default. pi's `SECTIONS` table already carries all 13 keys (the four rule slots sit in it with empty defaults), and `buildAcpSystemPrompt` fills rule slots from `prompts` *before* applying section overrides (`src/system-prompt.ts:93-103`). So the "application path untouched" claim holds, and the documented precedence (a `promptSections` string wins over a gated `prompts` override) follows directly from application order.
- Confirmed pinned 0.0.66 ships `leanPack...pi.promptSections.howToCompress === null`, and freshly cloned kernel master (now at `db03bf8`, i.e. *ahead* of the `c843f48` you tested — it picked up #271 since) ships the condensed `HOW TO COMPRESS (condensed)` string at `src/packs.ts:197`.

**Reproduced verification (both boundaries):**
- Pinned 0.0.66: `npm run typecheck && npm test && npm run build` → typecheck clean, 803 tests / 0 failures / 6 skipped, build OK.
- Kernel-master overlay (built locally, overlaid onto node_modules, restored afterwards — nothing committed): full suite green again, 803 tests / 0 failures. This exercises the *string* branch of the new version-tolerant assertions (`tests/prompt-pack.test.ts:57-59`, `:81-85`) that 0.0.66 exercises in its null branch — so both sides of the kernel boundary are covered, and the next kernel bump lands on a green tree.

**Test-quality notes (all good):**
- The junk-fixture swap in `tests/prompt-pack.test.ts:185` (`philosophy: "no"` → `tier3: 9`) was necessary, not cosmetic: under the widened contract the old fixture value would have been *kept*, breaking the expected `{ tools: null, whenToCompress: "keep" }`. Correct catch.
- New `tests/surface-config.test.ts:36-47` covers replace/remove/junk end-to-end through `buildAcpSystemPrompt`, including neighbor-section integrity.
- One minor fragility (not blocking): the else-branch `!text.includes("HOW TO COMPRESS")` at `tests/prompt-pack.test.ts:84` relies on no *other* lean section containing that phrase. It passes on both kernel sides today; if a future kernel adds such text, the assertion will need tightening. Not worth changing in this PR.

No issues to fix directly — nothing found. Merge is yours per house rules: https://github.com/ranxianglei/billion-context-pi/pull/427

中文摘要：独立复核了 kernel 侧 `applySectionOverrides` 契约（null=删 / string=替换，与键名无关）并在 0.0.66 与更新的 kernel master（db03bf8）两侧本地全量复跑（均 803 测试 0 失败），确认三态统一改动正确、测试版本容忍分支双向覆盖、文档无残留旧表述——可以合并。

PR #441, opened 2026-09-13
fix: guard kernelTextByCallId to tool-call cores only (#440)

### LINKED ISSUE #440: bug: kernelTextByCallId captures agent_message text incorrectly
## Bug: kernelTextByCallId incorrectly captures agent_message text as kernel output

**Version:** v0.1.69 (also present in v0.1.67)  
**File:** `src/messages.ts` line 270  
**Impact:** In Prime Agent, agent messaging tool text leaks into unrelated tool results

### Root Cause

```typescript
// src/messages.ts lines 268-271
const kernelTextByCallId = new Map<string, string>();
for (const core of coreOut) {
  if (core.toolCallId && core.text) kernelTextByCallId.set(core.toolCallId, core.text);
}
```

This captures ALL messages with `toolCallId` + `text`, but `agent_message` results also have both fields despite not being kernel output.

### Reproduction

1. Use Prime Agent with the `agent_message` tool
2. Send: `await agent_message.send("test", receiver_role='parent')`  
3. The message creates a core with `contentType: "agent-message"`, `toolCallId: "call_..."`, and `text: "test"`
4. Line 270 incorrectly captures it into `kernelTextByCallId`  
5. Subsequent tool results show "test" as their kernel output

### Proposed Fix

Add `contentType` check to filter only actual tool results:

```typescript
for (const core of coreOut) {
  if (core.contentType === "tool-call" && core.toolCallId && core.text) {
    kernelTextByCallId.set(core.toolCallId, core.text);
  }
}
```

This ensures only genuine tool-call results are captured, excluding agent messages.

I can submit a PR with this one-line fix + test case if helpful.

ranxianglei:
Fixes #440

## What this fixes

`kernelTextByCallId` in `coreOutToAgentMessages` captured **every** core message that had both a `toolCallId` and `text`, not just assistant tool-call cores. Tool-result cores (`contentType: "tool-result"`) share the same `toolCallId` as their call and carry the tool's *output* in `.text`, so they overwrote the call's argument text in the map.

That map feeds `syncToolCallArgs` → `compactedArgsFrom`, whose job is to sync kernel-compacted *arguments* back onto the reconstructed outbound tool-call blocks. When a tool result's text parses as a JSON object that differs from the original arguments, `compactedArgsFrom` replaced the reconstructed call's arguments with the parsed output. Net effect: the model sees fabricated/wrong arguments on its own prior tool calls.

## Reproduction (now a regression test)

Added a case in `tests/anchor-text-sync.test.ts`: a multi-call assistant message (`bash` + `grep`) plus their tool-result cores carrying JSON-object outputs. Before the fix the reconstructed `bash` arguments `{ command: "ls -la" }` were silently replaced by its own result `{ total: 5, files: ["a.txt"] }`; after the fix they are preserved.

## The fix

Restrict capture to genuine tool-call cores — the same guard `patchRefTag` already uses (src/messages.ts:377):

```ts
if (core.contentType === "tool-call" && core.toolCallId && core.text) {
  kernelTextByCallId.set(core.toolCallId, core.text);
}
```

## Scope notes

- Only the **multi-call** reconstruction path (`reconstructToolCallMessage`) consumes this map. The single-call path (`patchRefTag`) already used the correct per-core guard and was unaffected.
- Corruption triggers when a result's text parses to a JSON object. Plain-string results (like the literal `"test"` in the report) don't parse and were a no-op through `compactedArgsFrom`, but they still polluted the map — the allowlist removes them regardless of shape.
- The allowlist is robust to whatever `contentType` the offending core carries: the legitimate contributors are exactly the `tool-call` cores emitted by `projectMessage`.

## Verification

- `npm run typecheck` — clean
- `npm test` — 694 pass / 0 fail (3 pre-existing skips)
- `npm run build` — success


ranxianglei:

**Review of PR #441** (`fix: guard kernelTextByCallId to tool-call cores only`) — ✅ independently verified, ready to merge.

I checked out the PR head and re-derived every claim rather than trusting the description.

**Diff cleanliness (first gate) — clean.** Exactly 2 files, every line on-purpose:
- `src/messages.ts` (+3/−1): the one-line guard
- `tests/anchor-text-sync.test.ts` (+29): the regression test

No lockfile churn, no whitespace noise, no stray files.

**Correctness — reproduced, not assumed.** I temporarily reverted *only* the guard (pre-fix `src/messages.ts:244`) and ran the new test → it fails: `bash args must not be replaced by its result output` (the reconstructed `{ command: "ls -la" }` was silently swapped for its own result `{ total: 5, files: ["a.txt"] }`). Restored the fix → passes. So the test genuinely catches #440; it is **not vacuous**.

**Root cause vs symptom.** Matches the #440 triage. The map is keyed by `toolCallId`, and non-`tool-call` cores share that id while carrying *output* in `.text`. When that text parses as a JSON object, `compactedArgsFrom` replaces the reconstructed call's arguments with the parsed output → fabricated arguments shown to the model on its own prior calls. The fix targets the source (which cores may contribute), not the downstream symptom. Only the multi-call path (`reconstructToolCallMessage`) consumes this map; the single-call path (`patchRefTag`, `src/messages.ts:377`) already used the same per-core guard and was unaffected.

**Why an allowlist over a blocklist.** `contentType === "tool-call"` selects exactly the set that legitimately contributes and excludes every colliding core *regardless of its actual `contentType`* — including the `"agent-message"` core named in the #440 report (this adapter's `projectMessage` normalizes such results to `"tool-result"`, but the guard holds either way). It also aligns the multi-call path with the guard already proven in production at `src/messages.ts:377`. Minimal, idiomatic, robust.

**CI (run on PR head `454cbf8`):**
- `npm run typecheck` — clean
- `npm test` — 694 pass / 0 fail (3 pre-existing skips)
- `npm run build` — success

**Verdict:** minimal, correct at the root cause, well-tested, diff-clean. No changes requested; nothing to fix on the branch. Merging is left to you (human-only per our rules). https://github.com/ranxianglei/billion-context-pi/pull/441

中文摘要：已独立复核——改动干净(仅 2 文件、每行都服务于修复),回归测试非空(去掉那行过滤即可复现"参数被工具输出覆盖"),根因定位准确(按 `contentType === "tool-call"` 白名单收敛到真正的 tool-call 核,并顺带对齐单调用路径的既有守卫),typecheck/test/build 全绿,可以合并。

PR #472, opened 2026-09-19
fix(#471): re-append host-injected live-only messages dropped by Pi-host rebuild

### LINKED ISSUE #471: [bug] Pi host: host-only live messages are dropped by stateFor() (pi-web 生成标题变成对话残句)
# [bug] Pi host: messages that exist only in the live array are dropped by `stateFor()` — breaks every host-injected instruction (pi-web "Generate title" writes conversational sentences as titles)

**中文摘要**：在 Pi 宿主上 `stateFor()` 只从持久化 entries 重建请求，**宿主注入到 live 消息数组、但不写进会话文件的消息会被静默丢弃**。pi-web 的 `POST /api/sessions/<id>/auto-name`（Generate title）正是这样注入标题指令的 → 模型看不到指令，看到的是上一轮的结尾，于是「接着说上一轮的话」，pi-web 把这句话当标题存下（例：「没有待办动作了：项目文件、临时产物、僵尸进程和残留端口都已清理完毕……需要新东西时喊我」）。已在 0.1.71 复现；附最小复现脚本 + 已在本机验证的补丁。

---

## Environment

- `billion-context-pi@0.1.71` (npm), Windows 11, Node 22.23.2
- host: **pi** (`@earendil-works/pi-coding-agent`) driven by **pi-web** (`@agegr/pi-web`)
- `master` checked 2026-09-19: the gate is still there — `src/runtime.ts:595`
  `if (!isPiHost(sm) && liveMessages && liveMessages.length > 0) {`

## Symptom

pi-web's **Generate title** button normally produces a topic title. With ACP loaded, it produced the *next assistant sentence of the old conversation* instead, e.g. for a session about a Tetris page:

| session (persisted msgs) | title before | title after the fix |
| --- | --- | --- |
| 446 | `全部清完，无待办、无后台任务、无残留进程和端口。需要新东西再叫我` | `用技能和智能体做一个网页俄罗斯方块` |
| 446 | `没有待办动作了：项目文件、临时产物、僵尸进程和残留端口都已清理完毕……` | (同上) |
| 51 | `没别的事就到这里 —— 问题已修复，无需再操作。有需要随时叫我。` | `Apex 启动报 30005 修复：重启清除卡死的 EAC 驱动` |

The title text is the model's reply to *nothing* — it never saw the title request.

## Root cause

`src/runtime.ts:560` `stateFor(ctx, liveMessages)` — on a Pi host the live array is deliberately ignored:

```ts
// omp fires the context event BEFORE the current user message is persisted … pi appends user
// messages to the session before the LLM call, so its buildContextEntries() is always current.
// Merge event.messages (the exact messages about to be sent, including the not-yet-persisted
// tail) with the persisted branch records on the omp path only.        // src/runtime.ts:587-594
if (!isPiHost(sm) && liveMessages && liveMessages.length > 0) {          // src/runtime.ts:595
  const merged = mergeLiveEntries(entries, liveMessages, state, origins);
  … return { state, coreMessages: entriesToCoreMessages(merged), entries: merged };
}
const coreMessages = entriesToCoreMessages(entries);                      // Pi path
```

The assumption “pi persists the current user message before the call, therefore entries are always current” holds for **pi's own** turns. It does not hold for a host that injects context it never persists — and `pi-web` does exactly that:

`app/api/sessions/[id]/auto-name/route.ts` copies the session state into a fresh Agent and appends its instruction to the **live messages only** (either into the last user message, or as one extra new user message). It never writes a `message` entry, so from ACP's point of view it never existed → the rebuilt request stops at the previous assistant message → the model continues that message → pi-web stores the sentence as the title.

Normal turns are unaffected, which is why this stays invisible until a host does a "sidecar" LLM call (title, summary, tags, classification…). Any such call loses whatever the host injected; if the injected content is *data* rather than an instruction, it is silently lost with no visible symptom.

ACP's own log shows the built request count matching the entries-derived count, not the live count (`[turn] … inMsgs=64 outMsgs=64` for a session with 51 persisted messages + 1 injected).

## Minimal repro

`repro-live-only-tail.mjs` (self-contained, ~70 lines) builds a 2-message session, appends a host-only user message to the live array exactly like pi-web, fires the extension's `context` hook, and checks whether the instruction survives:

```text
$ node repro-live-only-tail.mjs ./bundle-pristine.mjs     # published 0.1.71
bundle:                    bundle-pristine.mjs
live messages:             3 (persisted 2 + 1 host-only)
stateFor() rebuild:        2 core messages, instruction delivered: false
context hook output:       2 messages, instruction delivered: false
RESULT: BUG - the host-only instruction was dropped; the model answers the previous turn and pi-web stores that as the title
exit=1

$ node repro-live-only-tail.mjs ./bundle-patched.mjs      # with the patch below
context hook output:       3 messages, instruction delivered: true
RESULT: OK  - the host-only instruction reached the model
exit=0
```

(`bundle-*.mjs` = `dist/index.js` + a few appended `export { … }` names, produced by `make-repro-bundles.mjs`; the repro itself only uses `createAcpExtension`, `createRuntime`, `SessionManager`, `createAgentSession`.)

## Suggested fix (works here, applied in the context hook)

Keep the Pi entries rebuild as-is (it is what keeps compression refs stable) and **re-append what only the live array carries**, right before the hook returns:

```js
// src/index.ts, in the `context` handler, just before `return { messages: rebuilt }`
const patchTail = liveOnlyTail(entries, event.messages);
if (patchTail) {
  rebuilt = rebuilt.concat(patchTail);
  logInfo("runtime", { sid, event: "live-only-tail-appended", tail: patchTail.length, outMsgs: rebuilt.length });
}
```

with detection that is structural, cheap and conservative (~0.3 ms on a 450-message session, one pass):

```ts
function patchEntryAsMessage(entry) {                      // persisted -> live-comparable shape
  return entry.type === "message"
    ? entry.message
    : { role: "custom", customType: entry.customType, content: entry.content };
}
function liveOnlyTail(entries, live) {
  const persisted = entries.filter((e) => e.type === "message" || e.type === "custom_message").map(patchEntryAsMessage);
  const max = Math.min(persisted.length, live.length);
  let i = 0;
  while (i < max && weakMessageSig(persisted[i]) === weakMessageSig(live[i])) i++;   // "role|customType|toolCallId|shape"
  if (i === max) {                                                                    // live is a strict extension
    const tail = live.slice(max);
    return tail.length > 0 && tail.length <= MAX_LIVE_ONLY_TAIL ? tail : null;         // MAX = 8
  }
  // pi-web can also append its instruction INTO the last user message: resend just the added text
  if (i === max - 1 && i === persisted.length - 1 && i === live.length - 1 && live[i].role === "user") {
    const before = patchTextOf(persisted[i]), after = patchTextOf(live[i]);
    if (before && after && after.startsWith(before)) {
      const suffix = after.slice(before.length).trim();
      if (suffix) return [{ role: "user", content: [{ type: "text", text: suffix }] }];
    }
  }
  return null;                                                                        // anything else: unchanged behaviour
}
```

Two traps worth knowing if you implement this differently:

1. **`custom_message` entries must be part of the alignment.** My first attempt aligned the live array against `entries.filter(e => e.type === "message")` only. Any session that ever received a `<background-task-notification>` (a `custom_message` entry, which Pi exposes to the model as `role: "custom"`) then lost its alignment run and the fix silently never fired — confirmed on a 446-message session (`[turn] inMsgs=552 outMsgs=109`, unchanged), while the same code fixed a 51-message session.
2. **Never re-derive the whole array from `merged` entries** for this case. Routing the turn through `mergeLiveEntries` (the existing omp path) also fixes the delivery, but on a Pi host it re-ids/ref-churns the live tail — the same area as **#459** ("live 尾部 id/ref 不稳定导致 compress 区间构造失败"). Appending the tail after the rebuild touches neither refs nor blocks.

If you would rather have one code path, the alternative is to relax `src/runtime.ts:595` to also merge when the live array contains messages the entries do not have (e.g. `live.length > entries.length` or the same `liveOnlyTail() !== null` test). I did **not** take that route because of #459 — maintainer's call.

## Verification of the patch (0.1.71 + local bundle patch)

- Unit + real-session regression suite: **23/23 pass** — aligned turn → `null`; appended prompt → the prompt; custom-message session → detected (regression for trap 1); "instruction inside the last user message" → only the added text; conservative rejects for shorter/中间 divergence/assistant-last divergence/non-prefix text/tail > 8; three real sessions (51/192/446 messages): aligned → `null`, live + prompt → `[prompt]`.
- Performance (450 messages): aligned pass **0.319 ms**/call, with tail **0.998 ms**/call.
- End to end through pi-web after restarting it: `event=live-only-tail mode=append live=449 persisted=448 tail=1` + `event=live-only-tail-appended tail=1`; the two sessions above got real titles.
- Standing cost when no host injects anything: one structural pass per turn (~0.3 ms), no behaviour change (tail = `null`, function returns early).

Happy to send this as a PR against `src/` (helpers + the 6-line hook insert + a test) if you want it in-tree — tell me where you'd prefer the detection to live (`runtime.ts` vs the hook) and I'll match the shape.

ranxianglei:
## Summary

Fixes #471 — on a Pi host, `stateFor()` builds the context purely from persisted session entries, so messages a host injects into the live array (`event.messages`) without persisting an entry are dropped. pi-web's "Generate title" relies on exactly that injection, and ends up storing a conversation fragment as the session title.

### Root cause
On a **Pi host** the else-path of the gate in `src/runtime.ts` runs `entriesToCoreMessages(entries)` and never consults `event.messages`; `mergeLiveEntries` is only used on non-Pi hosts. pi-web's auto-name route appends its instruction to `event.messages` **only** (a new trailing user message, or a text suffix into the final user message) without writing a session entry — so the injected prompt is lost, the model has nothing new to answer, continues the previous assistant turn, and pi-web captures that sentence as the title.

### Fix
New module `src/live-only-tail.ts` exports `liveOnlyTail(entries, live): AgentMessage[] | null`:
- Aligns persisted entries (`message` + `custom_message`) against `live` positionally with a weak content signature (role / customType / toolCallId + length + head/tail text), invariant to timestamps and ref tags.
- Returns the trailing **live-only tail** when present; the suffix-into-final-user-message case returns just the added text.
- Conservative contract: any unexpected shape or mid-sequence divergence → `null` → byte-for-byte unchanged behaviour. Tail capped at 8 messages.

Wired into the context handler in `src/index.ts` right before the return: when non-null, the tail is appended to `rebuilt`.

### Why append-after-rebuild (and not relaxing the Pi-host gate)
Relaxing the gate would route live messages through `mergeLiveEntries` — the volatile `live-N` id path behind the **#459** fork-host ref churn. This fix is **append-only** after the persisted rebuild: it touches neither refs nor blocks, so it stays orthogonal to #459. On non-Pi hosts the live array is already merged into entries, so the call returns `null` (no double-append); normal Pi turns align byte-for-byte → `null`.

Trap handled: pi projects every `custom_message` into the live array as `role:"custom"` (not `user`), so alignment maps `custom_message` → `{ role: "custom" }` — otherwise a `<background-task-notification>` would break the prefix walk.

### Verification (on current master, acp-kernel 0.0.75)
- `npm run typecheck` — pass
- `npm run build` — pass (self-contained ~835 KB bundle)
- New `tests/live-only-tail.test.ts` — **12/12 pass**: aligned turn → null; one/two appended trailing messages returned in order; tail over cap → null; live shorter than persisted → null; middle divergence → null; suffix-into-final-user-message returns only the added text (and the modified-not-last sub-case → null); non-prefix change on final user msg → null; final-msg mod that is assistant → null; `custom_message` session stays aligned and still detects an appended tail; empty live → null.
- Full-suite A/B: the failing set is **byte-for-byte identical** with and without this change on current master (all pre-existing environment failures) — no regression; the 12 new tests land on the passing side.

### 中文摘要
修复了 Pi 宿主下 `stateFor()` 只从已持久化条目重建上下文、从而丢弃宿主注入到 `event.messages` 的「仅存活消息」(pi-web 生成标题把指令只追加进 live 数组、未落盘)导致标题变成对话残句的问题。新增 `liveOnlyTail()`:在持久化重建之后按位置对齐,仅追加末尾的 live-only 尾巴;`null` = 行为逐字节不变,非 Pi 宿主本就合并了 live 故为 no-op,纯追加不触碰 refs/blocks,与 #459 正交。typecheck / build / 单测(12/12)全绿;全量测试失败集与改动前逐条一致(均为既有环境问题),无回归。

Closes #471


ranxianglei:

[bot] 🏷 收到，这是已完成的 PR（ework-agent-pr 标记），不重做。正在拉取分支 `2026-09-19_live-only-tail-pi-host` 核对代码、跑 typecheck/test/build，稍后给出结论。

ranxianglei:

[bot] 🏷 ## Review 结论：可合并（PR 合并按规矩由人工执行）

**查重与定位**：非重复项；针对 #471（Pi 宿主下宿主注入的 live-only 消息被 `stateFor()` 丢弃）。已验证根因属实：`src/runtime.ts:595-604` 的 gate 确认 Pi 宿主走 `entriesToCoreMessages(entries)`、从不读 `event.messages`，`mergeLiveEntries` 仅限非 Pi 路径。用户描述的是根本原因本身，方案（重建后纯追加）打到核心且选对了机制——放宽 gate 会把 live 消息送进 `live-N` 易失 id 路径（#459 ref churn），本 fix 在 `processTurn` 之后追加，不触碰 refs/blocks/token 快照，正交性成立。

**核对内容（全部实跑，非纸面）**：
- **diff 整洁度** ✅：3 文件 +265/−0，纯新增（`src/live-only-tail.ts` 新模块、`src/index.ts:694-703` 接线、`tests/live-only-tail.test.ts`），无锁文件/无关文件/空格重排；单 commit；分支基于当前 master（`59a652e`），无 stale base。
- **代码审查** ✅：`liveOnlyTail()` 保守契约成立——弱签名（role/customType/toolCallId+长度+首尾 32 字）对时间戳/ref tag 不变量；任何不确定形态一律 `null`→逐字节不变；尾长上限 8；`custom_message→{role:"custom"}` 映射正确处理 pi 投影陷阱（测试 trap #1 覆盖）；suffix 恢复仅在等长且仅末位分歧且为 user 严格后缀时触发，只返回增量文本作为新用户消息——前缀逐字节保持，**prefix cache 不被破坏**（§7.3 第一正确性属性）。无 `as any`/`@ts-ignore`。
- **宿主路径全覆盖**（§7.2）✅：Pi 正常轮次 live⊆entries→对齐→null；Pi+pi-web 标题注入→追加尾/suffix；非 Pi 宿主 `mergeLiveEntries` 已合并→对齐或保守 null，无双重追加；OMP 拒绝/proxy stand-down 路径在此之前返回，不受影响。
- **证据** ✅：typecheck 通过；build 通过（自包含 834.75 KB，与 PR 声称 ~835 KB 一致）；新测试 12/12 通过；全量 A/B：master 833 tests / 82 fail vs PR 845 tests / 82 fail，**失败集逐条 diff 完全一致**（均为既有环境失败），12 个新测试全落在通过侧——无回归。GitHub CI 全绿（test ubuntu/windows ×22/24、pr-validation、e2e ×2、build-artifact、docker）。

**一处小观察（非阻塞）**：追加的 tail 不计入 `tokenCount` 计量（计量在 `processTurn` 前完成）。影响面与既有 nudge 注入同类（同为未计量的合成追加），且 tail ≤8 条、通常仅一条短指令，provider 实际 usage 锚定校准兜底——可接受，不必为此改。

**Auto-merge gate（§7.4）**：虽然满足单模块/回归测试/全绿/pure fix:/引用 issue，但接线点在 `src/index.ts` 事件处理——§7.2 点名的热文件，属 rule 4 保留给人工的范畴。**请人工合并**：https://github.com/ranxianglei/billion-context-pi/pull/472

---
修复了 Pi 宿主下 `stateFor()` 只从持久化条目重建上下文、丢弃宿主注入到 `event.messages` 的 live-only 消息（pi-web 生成标题因此变成对话残句）的问题；方案为持久化重建后按位置对齐并纯追加末尾 live-only 尾巴，保守契约下任何不确定形态都回退到逐字节不变行为，prefix cache 与 #459 均不受影响；本地 typecheck/build/单测(12/12)/全量 A/B 失败集一致 + GitHub CI 全绿，可以合并。

PR #481, opened 2026-09-20
fix(#480): re-wrap bare-object compress content strings before kernel parse

### LINKED ISSUE #480: lean模式  Invalid compress content (content-not-array): got a JSON-encoded string (non-strict-tool providers stringify array arguments); the JSON failed to parse: Unexpected non-whitespace character
compress                                                                                                                                                                                     
 Invalid compress content (content-not-array): got a JSON-encoded string (non-strict-tool providers stringify array arguments); the JSON failed to parse: Unexpected non-whitespace character 
 after JSON at position 297 (line 1 column 298). Fix the malformed JSON (e.g. a missing closing brace or quote) and retry.

ranxianglei:
## Problem (issue #480)

In lean mode the `compress` call failed with:

```
Invalid compress content (content-not-array): got a JSON-encoded string (non-strict-tool providers stringify array arguments); the JSON failed to parse: Unexpected non-whitespace character after JSON at position 297 (line 1 column 298). Fix the malformed JSON (e.g. a missing closing brace or quote) and retry.
```

The diagnostic is not actionable: the payload is not "malformed JSON" in any sense the model can fix by retrying — it is structurally bracket-less, and every retry fails identically.

## Root cause

Small models in non-strict tool-call mode sometimes emit `content` as a string of **bare range objects without the wrapping array brackets** — one object (`{"startId":...}`) or several concatenated (`{...}{...}`, missing the outer `[]` and/or commas between entries).

Reproduced byte-for-byte: with the first object exactly 297 chars long and a second concatenated after it, strict `JSON.parse` fails with exactly `Unexpected non-whitespace character after JSON at position 297`, and `normalizeRanges` returned exactly the #480 error text. The kernel's lenient parser cannot handle this shape: `tryParseLenient` is whole-string (trailing garbage is fatal), and its salvage path is anchored on a literal `"content": [` array bracket, so bare objects never match → `content-not-array`. Even a single *clean* bare object fails today (the kernel only accepts arrays from the `content` field).

## Fix

Host-side deterministic repair in `src/compress-tool.ts` (same pattern as the existing #253 `tailRepair`):

- new `arrayWrapRepair(s)`: scans the content string for balanced `{...}` segments at depth 0 (string/escape-aware; stray closers resync instead of aborting), keeps only those that validate as range specs (same field aliases the kernel's `validateEntry` accepts: `startId|startRef|messageId`, `endId|endRef|messageId`, non-empty `summary`), and returns them as a proper JSON array string.
- `repairBareRangeObjects(args)` applies it to string `content` inside `normalizeRanges`, **before** `repairContentTail` (the two repairs cover mutually exclusive shapes).
- Conservative by construction: valid array strings (`[` first char), double-stringified arrays (one quoted token), line-form entries (no braces), whole-args-nested-in-content (`{"content":[...]}` — not range-shaped), and pure garbage all yield zero candidates → args pass through untouched and the existing accurate diagnostics still apply.

Why host-side rather than acp-kernel: the kernel's salvage design is `[`-anchored on purpose; this adapter already carries model-malformation repairs for its own failure modes (#253 missing-brace tail), and this keeps the fix shippable without a cross-repo release cycle. The same repair could later be upstreamed to acp-kernel for other hosts (opencode-acp) if desired.

## Verification

- New `tests/compress-bare-objects.test.ts` (13 tests): the exact #480 payload (position-297 signature asserted as a precondition), single clean bare object, trailing/leading garbage, per-object topic preservation, braces/quotes/escapes inside summaries, non-range-object filtering, no-false-positive matrix, and regression guards for the unchanged error paths. All pass.
- `npm run typecheck` ✓ · `npm run build` ✓
- Full suite: failure set is **byte-identical** with and without this change (79 pre-existing sandbox failures, all `EACCES ... /tmp/...` from this environment's read-only `/tmp` mount; they write session fixtures to `/tmp`). Zero regressions introduced.

Closes #480


ranxianglei:

[bot] 🏷 Reviewed PR #481 (fix for #480). I pulled the branch, reproduced the bug, verified the fix end-to-end, and checked CI. **Verdict: code is correct, minimal, well-tested, and conservative — approve.** One pre-merge step: rebase onto current `master` (details below). I am not merging (human-only).

## What I verified

- **Root cause reproduced byte-for-byte.** On base, feeding the exact #480 payload (first object exactly 297 chars + a second concatenated) through `normalizeRanges` returns precisely the reported, non-actionable error: `Invalid compress content (content-not-array): ... Unexpected non-whitespace character after JSON at position 297`. Confirmed real, not a misread of the symptom.
- **Fix works.** Same payload on the PR branch returns `array(2)` — both ranges recovered (`m00001..m00040`, `m00041..m00060`).
- **New tests:** `tests/compress-bare-objects.test.ts` → **13/13 pass** locally.
- **`npm run typecheck` ✓ · `npm run build` ✓** (clean bundle).
- **CI fully green** on the PR head: test matrix (ubuntu/windows × node 22/24), `pr-validation`, both `e2e` jobs, `build-artifact`, `docker` — all `success`.

## Code review (`src/compress-tool.ts`)

- `repairBareRangeObjects` runs **before** `repairContentTail` in `normalizeRanges`. The two repairs are genuinely mutually exclusive: a re-wrapped payload ends in `}]`, so `tailRepair`'s `body.endsWith('"')` guard no-ops; and any input with a leading `[` puts every `{` at depth ≥ 1, so `extractTopLevelObjects` yields zero candidates and passes through untouched for the tail repair. No interference.
- `extractTopLevelObjects` is string/escape-aware and resyncs on stray closers instead of aborting — robust to leading/trailing garbage and nested braces.
- **Field aliases match the kernel exactly.** Cross-checked against `acp-kernel`'s `validateEntry` (`src/parse-compress-input.ts`): `startRef|startId|messageId` / `endRef|endId|messageId` + required `summary`. The code comment's claim holds.
- Conservative by construction: valid arrays, double-stringified arrays, whole-args-nested-in-content, line-form entries, and pure garbage all yield zero candidates → args pass through unchanged and the existing accurate diagnostics still apply.

## Pre-merge step — rebase onto latest `master`

The PR branched off `dad487a`; `master` has since advanced. Two things to note:
- `master` also touched `compress-tool.ts`, but only in a **non-overlapping region** (top-of-file imports + a `#454` fork-ref-drift log block inside `handleCompress`). The PR's edits (`normalizeRanges` + new helpers after `tailRepair`) don't collide, so the rebase should apply cleanly with no conflicts.
- `master` bumped `acp-kernel` **0.0.69 → 0.0.78**. The PR was validated against 0.0.69; kernel's `validateEntry` aliases are unchanged on kernel master, so the fix remains valid — but please rebase and let CI re-run against 0.0.78 before merging.

## Minor, non-blocking observation

`isRangeLikeObject` requires `summary.length > 0`, whereas the kernel's `validateEntry` only rejects an *undefined* summary. So a bare object with an empty-string summary would not be repaired and would fall back to the old diagnostic instead of being salvaged. Extremely unlikely in practice (empty summaries are meaningless) and degrades gracefully — no crash, just the pre-existing path. Not worth blocking over; flagging for completeness.

## Duplicate screen
This is the designated fix for #480; no separate open PR covers the same change.

---
**中文摘要**:修复了 lean 模式下小模型把 `compress` 的 `content` 输出成无外层数组括号的裸对象、导致内核解析报误导性 `content-not-array`(position 297)错误的问题;在 `src/compress-tool.ts` 新增 `arrayWrapRepair` 在交给内核前把顶层裸对象重新包成合法数组,字段别名与内核 `validateEntry` 完全一致,13 个新测试全绿、本地 typecheck/build 通过、CI 全矩阵通过——逻辑正确且最小化,**可合并**(合并前请先 rebase 到最新 master 并让 CI 针对 acp-kernel 0.0.78 重跑)。按规则我不执行合并。

ranxianglei (commented):
Independent review — **approve, mergeable**.

Verified on the rebased head `1482342`:
- Root cause independently confirmed: bare-object content strings (`{...}{...}` without the wrapping `[]`) never match the kernel's `[`-anchored salvage path, and `tryParseLenient` fails whole-string on trailing content — so the "Fix the malformed JSON" hint is unactionable for the model.
- `arrayWrapRepair` is sound and well-bounded: string-escape-aware balanced-brace scanner, stray-closer resync, range-like validation mirrors the parser's accepted aliases (startRef/messageId etc.), proper-array strings yield no top-level objects (no interference with the happy path or `repairContentTail`, which composes after it). Non-repairable shapes still fall through to the accurate diagnostics.
- Local gates on my side: typecheck ✓, full suite **861 pass / 2 fail — the 2 failures (#415 delegate-detection tests) are pre-existing on master too in this environment** (848 pass / same 2 fail), so zero regressions; all 13 new tests pass; build ✓.

One note for the record: the kernel-side counterpart (making `tryParseLenient` salvage bracket-less objects) would also be reasonable upstream hardening, but the host-side repair is the correct immediate fix since the kernel pin moves slower.

PR #424, opened 2026-09-13
feat(compress): list remaining compressible ranges in the success panel

ranxianglei:
Fixes #420

## What
The successful `compress` result was built from header + warnings + errors only; the post-compression compressible-range snapshot (`compressibleSnapshotText`) was emitted on the rejection paths but not on success. A model wanting to plan a second batch in the same turn had to call `acp_status` first — the exact post-compress round-trip the recall-alignment work (acp-kernel#272 / PR #275) is eliminating.

This appends the post-compression snapshot on the success path, built from the already-computed `afterTurn` nudge (`src/compress-tool.ts`):

```ts
if (blocksCreated > 0) {
  const afterRanges = viableRanges(afterTurn.nudge?.compressibleRanges ?? []);
  if (afterRanges.length > 0) {
    lines.push("Current compressible ranges (use these refs exactly as listed):\n" + formatRanges(afterRanges, []));
  }
}
```

Result shape after a successful fold:

```
▣ ACP | 61.1K → 13.7K tokens (~47.4K reclaimed, blocks: b3=m00001)
Current compressible ranges (use these refs exactly as listed):
m00002–m00003 (12.0K tokens, 2 msgs)
```

## Tradeoff decision (raised in #420)
Every successful compress now carries the range list (~100–300 tokens) when anything viable remains. Accepted over threshold-gating because: (1) the cost is paid only on successful folds — rare events that themselves reclaim far more than the list costs; (2) the kernel surface provides no usage-gated variant here (`nudge.compressibleRanges` is computed unconditionally), so gating would mean inventing a new config knob for marginal savings; (3) both rejection paths already carry the identical snapshot, so this is consistency, not novel bloat; (4) it lets acp-kernel PR #275's rule harden from "one sanctioned planning `acp_status`" to a full no-round-trip policy.

## Tests
- positive: 4×6000-CJK entries, fold m00001 → result carries the section with m00002/m00003 refs (e4 protected as last message);
- negative: 2×6000-CJK entries, fold m00001 → e2 fully protected → no section emitted.

## Verification
Rebased onto current master (`0ebecee`, v0.1.68 / acp-kernel 0.0.66). `npm run typecheck` ✓ · full suite 802 pass / 0 fail (two consecutive runs; one intermediate run showed 3 transient timing flakes not reproduced, unrelated to this change — no test in it touches the new output path deterministically) · `npm run build` ✓. Diff is 2 files / 59 insertions, every line on-purpose.

[...]

PR #485, opened 2026-09-20
fix(#420): compress success result carries the post-compression ranges snapshot

### LINKED ISSUE #420: compress success result omits remaining compressible ranges — same-turn follow-up compression forces an acp_status call
来源: ranxianglei/acp-kernel#272 (https://github.com/ranxianglei/acp-kernel/issues/272) 分析 lean post-compress verification hole 时发现

## Problem
The successful compress tool result is built from header + warnings + errors only (`src/compress-tool.ts:501-507`):

```
▣ ACP | X → Y tokens (~Z reclaimed, blocks: b3=m00008–m00057)
```

`compressibleSnapshotText` (`src/compress-tool.ts:271-277`, "Current compressible ranges (use these refs exactly as listed)") is emitted on the rejection paths (`deadRepeatRejectionText` :279, `cappedRejectionText` :291) but **not** on success. So a model that wants to compress a second batch in the same turn must call `acp_status` first — exactly the post-compress round-trip the recall-alignment work (#259 / acp-kernel#272) is trying to eliminate.

Note: acp-kernel#272's premise "the result panel lists remaining compressible ranges" is inaccurate for current master. The kernel rule fix there therefore states only what is true (no re-verifying your own fold; the result confirms the block spans) and keeps one planning `acp_status` call sanctioned until this lands.

## Proposal
Append the post-compression snapshot on the success path, built from the already-computed `afterTurn` nudge (`src/compress-tool.ts:458`):

```ts
if (blocksCreated > 0) {
  const afterRanges = viableRanges(afterTurn.nudge?.compressibleRanges ?? []);
  if (afterRanges.length > 0) lines.push("Current compressible ranges (use these refs exactly as listed):\n" + formatRanges(afterRanges, []));
}
```

**Tradeoff to decide:** every successful compress would then carry the range list (~100–300 tokens) even when no further compression is intended. Alternatives: emit only while context usage stays above a threshold, or keep the status quo and accept the planning call.

## Impact
- Removes the last *legitimate* reason for a same-turn `acp_status` after a successful compress.
- Makes future prompt wording ("the result already lists the remaining ranges") truthful, so the lean rule can harden further.

ranxianglei:
Closes #420.

A successful compress now appends the remaining compressible ranges (built from `afterTurn` over the applied state), so a same-turn follow-up compress has its refs without a planning `acp_status` call. Silent when nothing viable remains.

Pre-flight: typecheck 0, compress-tool 15/15, full suite 864/866 (the 2 failures are pre-existing master #415 env-dependent detection tests, fail on unmodified master too).


ranxianglei:

[bot] 🏷 On it — this is a `[PR]` pickup, so I'm verifying rather than redoing: pulling branch `2026-09-20_issue420-success-ranges` (d8e5e86), reviewing the diff, running typecheck/test/build locally, and checking CI. Note: #420 previously had PR #424 — checking its state to make sure we're not carrying two live PRs for one issue. Full verdict shortly.

ranxianglei:

[bot] 🏷 Verification complete — **PR is merge-ready** (merge itself stays with you, per repo rules).

**Diff review (2 files, +50/−0, every line on-purpose)**
- `src/compress-tool.ts:602-611` — success path appends the snapshot only when `blocksCreated > 0`, built from `afterTurn.nudge` (the unconditional `processTurn` over the *applied* state at :545), so the list reflects what actually remains post-fold, not the pre-compression view. It reuses `compressibleSnapshotText()` (:355-361) — the same helper both rejection paths use — so success/rejection panels are format-consistent; empty case returns the "No compressible ranges remain…" sentence, which the guard filters out → silent when nothing viable remains, exactly as the issue specifies.
- `tests/compress-tool.test.ts:370-407` — one test covering both directions: fold m00001 of three 6K-CJK entries → section present with the untouched `m00002` ref; fold the rest → section gone. Follows the file's existing conventions (`/tmp/pai-acp-*` statefile naming etc.).

**Evidence**
- CI on d8e5e86: all 9 check-runs green (test matrix ubuntu/windows × node 22/24, e2e ×2, build-artifact, pr-validation, docker). The PR's `mergeable_state: blocked` is branch protection awaiting human approval — not a CI failure.
- Local re-run on this branch: typecheck ✓ · build ✓ · full suite failure set **identical to unmodified master** in this sandbox (both fail 82 tests on `EACCES /tmp` — this daemon's `/tmp` is a read-only mount, so every integration-style test that writes its sidecar to `/tmp` fails here regardless of code; the only delta vs master is the new #420 test, which fails locally for the same environmental reason and passes in CI). Zero regressions introduced by this change.
- No load-bearing surface touched: no ref-tag rendering, token calibration, host detection, sidecar format, `update.ts`, or acp-kernel pin changes; tool-result text only (appended at context end — no prefix-cache impact). Meets the §7.4 auto-merge gate criteria for a pure single-module `fix:` with regression tests.

**Duplicate cleanup:** PR #424 (Sep 13, same fix, stale v0.1.68-era base, never merged) was still open — two live PRs for one issue. I closed it as a duplicate of this one with an explanatory comment; branch preserved. #485 is now the sole live PR for #420.

Non-blocking nit (no change requested): the emptiness guard string-prefix-matches the hardcoded "No compressible…" message rather than checking range count — fine today since both live in one file, but if that sentence ever gets reworded the guard would need updating too.

After merge: acp-kernel#272/#275 can harden its rule from "one sanctioned planning `acp_status`" to the full no-round-trip policy — flagged so that follow-up isn't lost.

中文摘要：已核对代码与 CI（9 项检查全绿），本地复跑确认与未改动 master 的失败集完全一致（沙箱 /tmp 只读导致的 82 个环境性失败，与本 PR 无关；新测试在 CI 通过），实现正确且未触碰任何关键面，可以合并；同时已把同修 #420 的陈旧重复 PR #424 关闭并留了说明。

ranxianglei:
**验证状态（#484 队列清理）**：typecheck ✅ · build ✅ · 全量测试 864+/871 pass——仅剩的 2 个失败是 `tests/integration.test.ts` 的 #415 stand-down 测试，根因是开发者本机 `~/.pi/acp.json`（`promptPack: "lean"`）泄漏进测试环境（HOME 未隔离），与本 PR 代码无关，已由 **#489**（test: isolate HOME）单独修复，两测在该 PR 下 35/35 绿。合并顺序：#489 先行或同批，无冲突（一个只动 tests/integration.test.ts，其余不动该文件）。

ranxianglei:

[bot] 🏷 **合并顺序确认：#489 先行或同批，无冲突 —— 已核实。** 另外在核对 #489 时发现并直接修掉了一个小 bug（已推分支，CI 全绿）。

**Merge order — verified**
- File-level disjointness holds: #489 touches only `tests/integration.test.ts` (+18/−3); #485 touches only `src/compress-tool.ts` + `tests/compress-tool.test.ts`. Both sit on current master (`72d8771`, v0.1.74) — master hasn't moved since, so no rebase needed for either.
- Root-cause claim checks out: the two #415 stand-down tests pass here (no lean-pack home config in this env), consistent with "fails only on dev boxes whose `~/.pi/acp.json` sets `compress.promptPack: \"lean\"`, green on CI's empty HOME". Also checked the isolation is per-fixture: `home = join(dirname(agentDir), "home")` where `agentDir` lives inside a fresh `mkdtemp` dir per fixture, so no cross-test home sharing, and fixture cleanup removes it.

**Bug found & fixed directly in #489 (commit `d5fdbac`, pushed to `2026-09-20_test-home-isolation`)**
The new `withAgentDir` teardown replaced the guarded restore with an unconditional `process.env.PI_CODING_AGENT_DIR = prev`. When the variable was previously unset, assigning `undefined` to `process.env` coerces it to the literal string `"undefined"` (verified: `env -u PI_CODING_AGENT_DIR node -e '…'` → prints `"undefined"`). That leaks into every later test in the process and corrupts agent-dir resolution at `src/setup-subagent-tools.ts:48` — ironic for a PR about env hygiene. Fix restores the original delete/set guard, matching the `HOME`/`USERPROFILE` guards added in the same hunk and the pattern already used in `tests/delegate-toggle.test.ts` / `tests/user-config.test.ts`.
- Local: typecheck ✓; `tests/integration.test.ts` results byte-identical pre/post fix (the 14 fails in *this* sandbox are the known read-only-`/tmp` EACCES noise, present on unmodified master too).
- Post-push CI on `d5fdbac`: all 9 checks green (test ubuntu/windows × node 22/24, e2e ×2, build-artifact, pr-validation, docker).

**Nit (fixed by editing the PR body, no code impact)**
#489's body cited `tests/enabled-config.test.ts` as the existing HOME-pinning precedent — that file doesn't exist; the real precedents are `tests/delegate-toggle.test.ts` / `tests/user-config.test.ts`. Body corrected in place.

**Bottom line:** #485 remains merge-ready as assessed earlier; land #489 first (or same batch) so the suite is clean everywhere, then #485. Merging stays yours per repo rules.

中文摘要：核对了 #489 与 #485 无文件重叠、可先行或同批合并；发现 #489 的 `withAgentDir` 收尾把未设置的 `PI_CODING_AGENT_DIR` 恢复成了字符串 `"undefined"`（会污染后续测试的目录解析），已直接在 PR 分支补上原守卫（commit d5fdbac），本地 typecheck+集成测试结果与修前一致、CI 9 项全绿，另把 PR 描述里不存在的测试文件名改成了实际先例文件。

PR #487, opened 2026-09-20
fix(#467): acp.json enabled switch survives hand-edited JSON shapes

ranxianglei:
Closes #467 (from billion-context#894).

Strict JSON.parse made BOM heads, unquoted keys, and trailing commas (the Windows-notepad shapes from the original report) all silently mean 'not disabled' — the opposite of the user's intent. The common shapes are now repaired (with a portability warning); non-boolean enabled values and unparseable files warn loudly instead of being swallowed. Missing files stay 'not disabled'.

Pre-flight: typecheck 0, enabled-config.test.ts 5/5, full suite 868/870 (2 pre-existing env failures on master).

PR #489, opened 2026-09-20
test: isolate HOME in withAgentDir fixtures

ranxianglei:
Found during the #484 queue triage while validating PR #487's full suite: the two #415 stand-down tests in `tests/integration.test.ts` failed on any developer box whose `~/.pi/acp.json` sets `compress.promptPack: "lean"`, while CI (empty HOME) stayed green.

Root cause: `withAgentDir` pinned only `PI_CODING_AGENT_DIR`. The prompt/config assembly (`before_agent_start` → `resolveActivePack` → `loadUserConfig`/`homedir()`) still read the runner's real home config. The lean pack swaps the `acpTags` section for its headerless short form, so the literal `"ACP TAGS"` assertions fail — masking as a fake #415 regression.

Fix: pin `HOME`/`USERPROFILE` to a fixture dir for the duration of `withAgentDir`, matching the HOME-pinning pattern already used in `tests/delegate-toggle.test.ts` / `tests/user-config.test.ts`.

Verified: 2 failures with a lean home config, 35/35 pass with isolation. typecheck + targeted suite green.


ranxianglei:

[bot] 🏷 Reviewing PR #489 now (branch `2026-09-20_test-home-isolation`, single commit `d08d675`, +18/-3 in `tests/integration.test.ts`). Plan: independently reproduce the lean-home-config failure on master, then validate the fix + full suite/typecheck. Will post findings here shortly.

ranxianglei:

[bot] 🏷 Reviewed — root cause confirmed, fix validated by reproduction, one small bug found in the PR and fixed directly on the branch.

**Reproduction (master, Node 22):**
- With a home config `~/.pi/acp.json = {"compress": {"promptPack": "lean"}}`: the `#415: project-scope pi-subagents → acp_delegate stands down` test fails with `'core ACP prompt still present'` — exactly the fake #415 regression described.
- With an empty HOME (CI-equivalent): same suite green. Mechanism matches your analysis precisely: `session_start` → `loadUserConfig(cwd)` (src/runtime.ts:541) reads `homedir()` → `~/.pi/acp.json`, then `before_agent_start` → `resolveActivePack` applies lean's headerless `acpTags` section.
- One nuance vs the issue text: on master I observe **exactly one** additional failure under a lean home, not two. The other two literal `"ACP TAGS"` assertions (the `delegate:false` / `delegate:{enabled:false}` tests) never call `session_start`, so user config is never loaded into the adapter and they stay immune. The lean pack (acp-kernel builtin) only overrides pi prompt sections, which is why the forceEnable/matrix #415 tests are also unaffected. Doesn't change the diagnosis or the fix.
- (Minor: the cited reference `tests/enabled-config.test.ts` exists only on the unmerged PR #487 branch, not master — the same HOME-pinning pattern is already in master's `tests/user-config.test.ts`.)

**Validation of the fix:** ran the full suite on the PR head under the lean home and diffed the failing-test set against an empty-home baseline: the #415 failure disappears, zero new failures anywhere.

**Bug found in the PR (fixed on the branch):** the new teardown restored `PI_CODING_AGENT_DIR` with a plain assignment (`process.env.PI_CODING_AGENT_DIR = prev`). When the variable was previously unset, Node's env setter coerces `undefined` to the **string `"undefined"`** (verified empirically), leaving a *truthy* value for every later test in the process. That's dangerous: `resolveAgentDir()` (src/setup-subagent-tools.ts:47-56) keys off truthiness and would then resolve the literal path `"undefined"` instead of falling back to `~/.pi/agent`. The old code had the delete/assign guard for exactly this reason.
- The guard restore (`if (prev === undefined) delete …`) was already landed on the branch as `d5fdbac` (pushed by another agent instance while I was validating — byte-identical to my fix, so no conflict).
- My commit `d1e6c0d` adds the missing regression test: asserts `PI_CODING_AGENT_DIR`/`HOME`/`USERPROFILE` are restored verbatim after `withAgentDir`; it fails against the old assignment (verified) and passes now.

**State:** typecheck clean, integration suite + full suite green under both lean and empty HOME on `d1e6c0d`. CI check runs for the new head were still in progress when I posted this — will confirm once they complete.

修复了测试夹具的 HOME 隔离问题并补上了 teardown 中 `PI_CODING_AGENT_DIR` 被污染成字符串 `"undefined"` 的守卫与回归测试，复现和全量验证均通过，CI 跑完确认后即可合并。

ranxianglei:

[bot] 🏷 CI update: all checks green on the current head `d1e6c0d` — test matrix (ubuntu/windows × Node 22/24), e2e (both OSes), docker, pr-validation, build-artifact all pass; PR is mergeable. Ready for you to merge when convenient (merges stay human-side).

CI 全部通过（含我补的回归测试提交），PR 可合并。

PR #495, opened 2026-09-21
honor AbortSignal in compress/decompress tools

ranxianglei:
## Background

Issue #494 reported two distinct problems:

1. **The 20-minute hang itself did not enter BCP.** Verified against the pi-agent-core agent loop (`agent-loop.js`): when an assistant turn ends with `stopReason: "aborted"`, the loop emits `turn_end`/`agent_end` and returns **without** calling `executeToolCalls()`. Both compress calls in the session have complete args but no `toolResult`, which matches this path exactly. Root cause is upstream of this repo: the provider stream (gpt-5.6-luna endpoint / network / adapter) stopped delivering events after the tool-call arguments were fully streamed, before the terminal `done` event, and no client-side stream timeout fired for 20 minutes. This PR does not address that layer (no raw stream capture exists to localize provider vs adapter).

2. **Independent BCP defect (fixed here):** the `compress` and `decompress` tools accepted the host `AbortSignal` but never checked it (`_signal` was unused). Once `handleCompress()`/`handleDecompress()` started executing, a user abort could not stop them — in-flight work, including state persistence (`runtime.save()`), could continue after the user pressed Esc.

## Changes

- New `src/abort.ts`: `assertNotAborted(signal)` throws `AbortError` (message `Operation aborted`, matching pi's own convention) at phase boundaries.
- Thread the signal through `execute(toolCallId, params, signal, onUpdate, ctx)` into both handlers; checkpoints at entry, after `await runtime.stateFor(ctx)`, and after `core.applyCompression` / before block-mode file writes.
- An aborted run now surfaces as an `isError` tool result instead of hanging or persisting half-applied state.
- New `tests/abort-signal.test.ts`: unit test for `assertNotAborted` plus e2e cases (pre-aborted compress rejects with AbortError and leaves the sidecar byte-identical; pre-aborted decompress rejects; live un-aborted signal still succeeds).

## Scope notes

- `search_context` / `acp_status` left unchanged: single fast phase, not named in the issue.
- `_onUpdate` progress reporting intentionally skipped: kernel phases are ms-scale synchronous work; emitting updates would only add transcript noise.
- No version bump (release-branch-only per project rules).

## Verification

- `npm run typecheck` clean; `npm run build` OK (dist self-contained, acp-kernel inlined)
- New tests 4/4 pass; full suite failures in my sandbox are all pre-existing environment artifacts (read-only `/tmp`) — verified against a stashed clean tree

Fixes the BCP-side defect reported in #494.


ranxianglei:

[bot] 🏷 Acknowledged. This is flagged `ework-agent-pr`, so I'm treating it as already-completed work: pulling the branch, reviewing the diff, running the local checks (typecheck / tests / build), and checking CI status before reporting a conclusion. Will follow up here with findings.

ranxianglei:

[bot] 🏷 **Verification complete — ready for human merge.** I pulled the branch, reviewed the diff, and ran the full local + CI verification. One note up front: the branch was force-updated mid-review (old head `4eca41f` based on stale master → new head `d205352`, now a single commit directly on current master `c323299`). All results below are for the current head.

**Diff review (clean, single-purpose, +159/−6 across 4 files):**
- `src/abort.ts`: `assertNotAborted(signal)` throws `AbortError` with message `"Operation aborted"` — matches pi's own convention; accepts `undefined`.
- `src/compress-tool.ts`: checkpoints at entry (`:343`), after `await runtime.stateFor(ctx)` (`:353`), and after `core.applyCompression` before either `runtime.save()` (`:426`, saves at `:431`/`:444`). An abort between apply and save discards the in-memory mutation — nothing half-applied gets persisted, as claimed.
- `src/decompress-tool.ts`: checkpoints at entry (`:222`), after `stateFor` (`:224`), and before the block-mode `mkdir`/`writeFile` (`:270`).
- Thrown `AbortError` propagates through the existing `logThrow` → rethrow path, so pi marks it `isError` — consistent with the tool's other failure modes.
- Scope notes hold up: `search_context`/`acp_status` untouched (single fast phase), no `_onUpdate` churn, no version bump (correct per release-branch-only rule).

**Local verification (head `d205352`):**
- `npm run typecheck` — clean
- New `tests/abort-signal.test.ts` — 4/4 pass (pre-aborted compress rejects with AbortError and leaves the sidecar byte-identical; pre-aborted decompress rejects; live signal still succeeds)
- Full suite: 902 tests, 808 pass, **88 fail — identical failure set (byte-for-byte diff) on clean master `c323299`**. All 88 are pre-existing sandbox artifacts: `EACCES ... lstat '/tmp/...'` because this environment mounts `/tmp` read-only while several older tests hardcode `/tmp` paths (e.g. `tests/decompress-tool.test.ts:63`). They pass in CI where `/tmp` is writable.
- `npm run build` — OK, `dist/index.js` self-contained (~888 KB, acp-kernel inlined)

**CI (GitHub Actions, head `d205352`):** all green — test (ubuntu/windows × node 22/24), e2e (ubuntu/windows), build-artifact, pr-validation, docker. `mergeable: true`, no conflicts.

**Two minor observations (no action required for this PR):**
1. Pre-existing test-infra wart: older tests hardcode `/tmp/...` instead of `os.tmpdir()`, so the suite can't run in read-only-`/tmp` environments (like this daemon's). The new test file does it right (`os.tmpdir()`). Worth a separate cleanup issue sometime.
2. The PR body says "Fixes the BCP-side defect reported in #494" — that phrasing won't trigger GitHub's auto-close. If you want #494 closed on merge, change it to `Fixes #494` (or add a line).

修了什么问题：compress/decompress 工具此前接收了宿主的 AbortSignal 却从不检查（`_signal` 未使用），用户 Esc 中止后工具仍会继续执行甚至持久化状态；本 PR 在阶段边界插入 `assertNotAborted` 检查点，中止时以 `AbortError`（isError）快速失败且不落盘半应用状态。为什么修：#494 中确认的独立 BCP 缺陷（20 分钟挂起本身属上游 provider 流问题，不在本 PR 范围）。验证结论：diff 干净、单模块、回归测试齐备，本地 typecheck/build/新测试全绿，全量套件失败集与干净 master 完全一致（均为沙箱 /tmp 只读导致的既有问题），CI 全绿且分支已 rebase 到最新 master —— **可以合并**（合并由人工执行）。

